PR #8, opened 2018-03-14
Update aws-resource-directoryservice-microsoftad.md

par-texx:
Change name of Ad to Enterprise Edition AD, and made note that a Standard Edition cannot be created through CF at this time.

*Issue #, if available:*

*Description of changes:*
Made comment about CF only being able to make an enterprise edition AD.


AWSRandall:
Thanks for sending us feedback! I'm checking on this and will get back to you as soon as we have a confirmation.

Thanks for using CloudFormation!

par-texx:
Thanks Randall.  If you have access, check support case 4821023561.

AWSRandall:
I'm double-checking with my editor on the right branding guidance. Otherwise, looks good!

Thanks again!

PR #12, opened 2018-03-14
StreamEncryption is not required for Kinesis Stream

PatMyron:
*Issue #, if available:*

*Description of changes:*


iann0036:
For these kinds of changes, will AWS also change the specification files to reflect it?

AWSRandall:
Hey, thanks for using CloudFormation! 

For these types of issues, I'm first looking at the source code to see if it matches what's in the docs. Then, I'm working with engineering to make sure everything is documented as it was intended to be implemented. So yes, there's a chance the specification file might change. 

I'll get back to you shortly.

AWSRandall:
The resource specification does not have `required="false"` (it does not have a `required` attribute). I'll file an issue with the engineering team and they can make the call to add the required="false" or not. 

I've fixed the docs and am pushing through the changes to our production servers.

Thanks again for spotting the issue!

PR #13, opened 2018-03-14
Document Arn property

hagbarddenstore:
The documentation only mentions the Arn property in one of the examples. It
should list the property under the return values header like other
CloudFormation resources do.


AWSRandall:
Nice catch! Thanks for sending us the feedback!

I also updated the GetAtt Function reference at https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getatt.html.

Merging and building to our prod servers. Thanks again!

PR #16, opened 2018-03-15
YAML examples do not requires the CloudFormation type to be enclosed …

kennyk65:
…in quotes.

*Issue #, if available:*

One of the great things about the YAML format is one does not have to spend so much time fussin' with commas, brackets, quotes, etc.  I've been using these samples a lot, and one thing I've noticed is all of the extra quotes on things like the CloudFormation type - they're not needed.  So I swept through and removed the double quotes enclosing the CloudFormation types from all of the YAML examples.  Sorry for the big PR, but I figure you all would want all the documentation to be consistent no matter what.

### DIFF
--- a/doc_source/aws-properties-as-group.md
+++ b/doc_source/aws-properties-as-group.md
@@ -47,7 +47,7 @@ To declare this entity in your AWS CloudFormation template, use the following sy
 ### YAML<a name="aws-resource-autoscaling-autoscalinggroup-syntax.yaml"></a>
 
 ```
-Type: "AWS::AutoScaling::AutoScalingGroup"
+Type: AWS::AutoScaling::AutoScalingGroup
 Properties:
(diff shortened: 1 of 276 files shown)


AWSRandall:
Hey, thanks so much for using CloudFormation! 

This is a good point; thanks for raising the issue and doing all this work! I'm going to need to double-check these for accuracy to make sure they work in CFN, so it's going to take some time to merge such a large PR. Thanks in advance for your patience.

kennyk65:
Love CloudFormation!  But now that I've gotten better I notice little imperfections in the provided samples.  In general, in YAML the double quotes on the values are **almost never needed.**

rjlohan:
This PR is accurate in that there is never a need to enquote the Type property. However in general we have to be aware of certain side-effects of YAML scalar interpretation causing incorrect type inference in some situations. This isn’t one of those cases but something to be mindful of before we unquote all the things. 😁

rachfop:
Thank you for your feedback, for any questions regarding this PR please see #520 .

PR #34, opened 2018-04-04
Moving Type above Properties in example YAML

keithsharp:
Pretty much all example CloudFormation YAML has the order Resource Name -> Type -> Properties, updating the example to match this.


ckesser:
@keithsharp This does help keep our docs more consistent, thanks for doing that. I'll merge the request.

PR #35, opened 2018-04-04
Fix indentation in example YAML.

keithsharp:
The list indicator (-) was indented with a tab rather than spaces in the example YAML which was making the indentation incorrect.


ckesser:
@keithsharp Thanks for catching this! Merging now.

PR #40, opened 2018-04-10
Update YAML example for AWS::Lambda::Function

tjpatter:
The Timeout property needs to be expressed as an integer (not a string).

*Issue #, if available:* 
Currently, the documentation's YAML syntax is incorrect regarding the Timeout property of the AWS::Lambda::Function type. If you were to copy / paste the YAML example and try to execute it you would receive errors on the Timeout property. The documentation even makes mention that this is to be expressed as an integer and not a string.

*Description of changes:*
Removed the double-quotes ("") surrounding the Timeout property in the example code.


ckesser:
@tjpatter Thanks for this! Merging now.

Issue #41, opened 2018-04-11
Inconsistent Update Behavior for DynamoDB SSESpecification on different pages


PatMyron:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-table.html#cfn-dynamodb-table-ssespecification

"Type: DynamoDB SSESpecification

Update requires: Some interruptions"

Following the link there takes me here:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dynamodb-table-ssespecification.html

Where there is only one property of DynamoDB SSESpecification and updating that property apparently requires replacement.

"Properties
SSEEnabled
Whether server-side encryption is enabled or not.

Required: Yes

Type: String

Update requires: Replacement"

PatMyron:
Merged https://github.com/awsdocs/aws-cloudformation-user-guide/pull/149

PR #42, opened 2018-04-11
Correcting inconsistent Update Behavior for DynamoDB SSESpecification

PatMyron:
… on different pages

The only property of DynamoDB SSESpecification says update will require replacement here:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dynamodb-table-ssespecification.html

*Issue #, if available:*

 https://github.com/awsdocs/aws-cloudformation-user-guide/issues/41

*Description of changes:*

### DIFF
--- a/doc_source/aws-resource-dynamodb-table.md
+++ b/doc_source/aws-resource-dynamodb-table.md
@@ -116,7 +116,7 @@ Throughput for the specified table, which consists of values for `ReadCapacityUn
 Specifies the settings to enable server\-side encryption\.  
 *Required: *No  
 *Type*: [DynamoDB SSESpecification](aws-properties-dynamodb-table-ssespecification.md)  
-*Update requires*: [Some interruptions](using-cfn-updating-stacks-update-behaviors.md#update-some-interrupt)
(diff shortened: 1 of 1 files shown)


PatMyron:
Not sure why editing files through the Github UI online results in that, but if you hide whitespace changes, you can see the only line I actually changed:

screen shot 2018-07-06 at 10 42 35 pm

ckesser:
@PatMyron Hey, thanks for submitting this. I checked the Resource Specification, which just complicated things further. The Resource Spec marks 'SSEEnabled' as requiring 'No Interruption'. I'm going to reach out to our product team and get back to you.

PatMyron:
"Encryption at rest can be enabled only when you are creating a new DynamoDB table. Currently, you can't enable encryption at rest on an existing table. After encryption at rest is enabled, it can't be disabled."
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/EncryptionAtRest.html

ckesser:
Superseded by #149

PR #50, opened 2018-04-17
Update using-cfn-waitcondition.md

kennyk65:
The WaitCondition doesn't really "pause the stack", nor does it help define when a resource is "really created" (that's the CreationPolicy).  Instead, it allows a dependency to be setup between resources where a specified number of signals must be received before the dependent resource can begin being created.  Resources unrelated to the wait condition are free to be created.

*Issue #, if available:*

*Description of changes:*


rachfop:
Thank you for your pull request. I was able to merge you revisions: however, the docs will be updated soon to ensure consistency with our style guide. The important part, which you've highlighted, is the accuracy of our documentation. Thanks.

PR #63, opened 2018-04-30
Replace short resource ID with long resource ID

walery:
Since there is no short resource ids for ec2 anymore they should not
be uses in docs as well.

*Issue #, if available:*

*Description of changes:*

See: https://aws.amazon.com/ec2/faqs/#longer-ids


ckesser:
Thanks for updating this example, we appreciate the contribution. Merging your request now.

PR #78, opened 2018-05-29
Update EncryptionConfiguration parent property

cariadeccleston:
`EncryptionConfiguration` is a property of `ReplicationDestination`, but the documentation implies it's a property of the top-level `AWS::S3::Bucket` resource type. I have updated the document so it references `ReplicationDestination` as the property's direct parent.


ckesser:
Thanks for this contribution! Merging the request.

PR #81, opened 2018-05-30
Update parameters-section-structure.md

indukuk:
List<Number> evaluates a number and the Ref function also evaluates to a number and Not to a string as was described in this section. So modified as Under

Removed - “ however, when you use the parameter elsewhere in your template \(for example, by using the `Ref` intrinsic function\), the parameter value becomes a list of strings\.

Added - For example, users could specify `80,20`, and a `Ref` would result in List of numbers `80 20`\. To convert this to string you can use 'Join'. For example, !Join [',',[!Ref Parameter]] would result in 80,20

*Issue #, if available:*

*Description of changes:*


rachfop:
Hello, your request was used to update the documents. You can see the changes [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html)

PR #85, opened 2018-06-05
Fixes the order of IAM resource types

KyleMarMc:
The `AWS::Inspector` resource types had been inserted in the wrong place - `AWS::IAM::UserToGroupAddition` was showing after them, rather than immediately after `AWS::IAM::User`.


ckesser:
Thank you for contributing! Merging your request.

PR #86, opened 2018-06-07
Fixed typo on call to mysql_error().

LaikaN57:
Two calls to mysql_error() were missing the "m". This commit fixes this and provides the correct error output if the db fails to connect.

mysql_error() documentation: http://php.net/manual/en/function.mysql-error.php

Lastly, I am not sure why github decided to mark line 868 as changed as the lines are exactly the same. I even checked to make sure that they were the same unicode character.


LaikaN57:
@rjlohan Looks like this was merged on 7/16/18 but not closed. Do you want me to close?

ckesser:
@LaikaN57 Thanks for this fix! And @rjlohan thanks for approving. Merging the request now.

PR #87, opened 2018-06-08
Update conditions-section-structure.md

kennyk65:
Apparently, you cannot refer to one condition from another condition to build complex conditions.  But the document doesn't mention this.

*Issue #, if available:*

*Description of changes:*


cmmeyer:
This is actually possible.  Here's an example from the VPC template used for many of our Quick Starts:

```
        "3AZCondition": {
            "Fn::Or": [
                {
                    "Fn::Equals": [
                        {
                            "Ref": "NumberOfAZs"
                        },
                        "3"
                    ]
                },
                {
                    "Condition": "4AZCondition"
                }
            ]
        },
```

PR #94, opened 2018-06-20
Include definition of ProvisioningArtifact

jeff-shep:
*Issue #, if available:*

*Description of changes:*
Updated aws-resource-servicecatalog-cloudformationprovisionedproduct.md

The term ProvisioningArtifact is misleading when working with Service Catalog. It should be clearer from this page that it is the same of product version. Either the update I suggest here or a link to https://docs.aws.amazon.com/servicecatalog/latest/dg/API_ProvisioningArtifact.html


ckesser:
@jeff-shep Thanks for this definition! Provisioning artifact is defined elsewhere in the guide, but not here, and I agree that it should be defined here, as well. Merging your request now.

PR #95, opened 2018-06-21
Clarification of EndpointURL attribute

uldall:
Made it clearer which value is returned from the EndpointURL attribute.

*Issue #, if available:*

*Description of changes:*
Changed the description of EndpointURL under Fn::GetAtt.


uldall:
I used the build-in text editor in GitHub to make this change. The normal diff shows that a lot of stuff changes. I think this is a bug in the GitHub text editor. The "Rich Diff" shows the change correctly.

PR #96, opened 2018-06-22
Bucket is specified by name in BucketPolicy

edenhochbaum:
Clarify that Bucket property of AWS::S3::BucketPolicy is specified by name (rather than, e.g., resource ARN)


ckesser:
Thanks for this! Merging the request now.

PR #97, opened 2018-06-23
"!Ref" is not shorter than "Ref"

phinnaeus:
*Issue #, if available:*

*Description of changes:*
Self explanatory. I'm guessing this was generated automatically, but it seems silly to even offer "!Ref" as an option when "Ref" is valid.

### DIFF
--- a/doc_source/intrinsic-function-reference-ref.md
+++ b/doc_source/intrinsic-function-reference-ref.md
@@ -29,7 +29,7 @@ Syntax for the full function name:
 Ref: logicalName
 ```
 
-Syntax for the short form:
+Syntax for the "short" form:
 
 ```
 !Ref logicalName


ckesser:
@phinnaeus You make a good point - the term ‘short form’ is misleading in terms of the length of ‘!Ref’. I spoke with my editor and learned that this use of quotes doesn’t align with our style conventions. 
I think it makes sense to document !Ref because the short form for intrinsic functions like Fn::Sub is in fact shorter than the full function name, and someone who typically uses the short form for these functions will likely use the same syntax for Ref regardless of length. I'm planning on leaving the documentation as-is to keep terminology consistent across intrinsic functions. I'm going to close out this request for now, but please feel free to let me know what your thoughts are; I’d be happy to work on another solution. Thanks for bringing attention to this!

PR #98, opened 2018-06-25
Add possible values to ReadWriteType

phinnaeus:
*Description of changes:* It's not clear what to actually put in your CFN template for non-default behavior for `ReadWriteType`. This change merely adds those options to the docs.


ckesser:
@phinnaeus Thanks for adding these valid values! Merging your request now.

PR #107, opened 2018-07-06
Remove BadgeEnabled from the example since Build badges are not supported for CodePipeline source

bundyfx:
"BadgeEnabled": "true" in the example template for yaml/json is a little misleading since its not currently supported with a pipeline source of **codepipeline** and ultimately fails when attempting to create the stack.

proposed change is to remove it from the example templates so that they will succeed "out of the box"


ckesser:
@bundyfx Thanks for this! Merging the request now.

PR #108, opened 2018-07-06
Fix ENUM value for SubscriptionType

acurley:
The ENUM for this resource is case sensitive.

*Issue #, if available:*

*Description of changes:*
Discovered the case sensitivity when provisioning a Budget resource.


ckesser:
Thanks for using CloudFormation and for reporting this issue! Merging your request.

PR #109, opened 2018-07-06
Add EnableNonSecurity to PatchBaseline Rule

acurley:
*Issue #, if available:*

*Description of changes:*
The EnableNonSecurity parameter is documented in the API but is seemingly left out of the CloudFormation documentation. Despite its undocumented status, it appears to actually be available through CFN as I have deployed baselines in my account using this in my templates.  As a result, I assume that it is there and just missing from the docs.


ckesser:
Thanks for this fix! Merging your request.

PR #110, opened 2018-07-09
Clarifying impact of change to LicenseModel on AWS::RDS::Instance.

rjlohan:
Current documentation incorrectly indicates that modifying the LicenseModel of an RDS::Instance will cause a replacement. This is both untrue, and frightening for customers. Have clarified the text to indicate that this change incurs **interruptions** _not_ **replacement**.


nishantcasey:
@rjlohan - I believe this should instead be "Updates with Some Interruption" or just simply "Some Interruptions" 

As the underlying EC2 host is replaced when migrating (say from BYOL to LicenseIncluded) as RDS will replace the underlying host. Even when DB Instance is MultiAZ - this can cause a few moments of downtime (and a few minutes for Single AZ instances)

nishantcasey:
I believe this should say "Some Interruptions" or "Updates with Some Interruptions" as the LicenseModel change will cause a few moments of downtime if the DBInstance is MultiAZ and longer if the instance is Single AZ.

rjlohan:
Modified PR to state 'Some interruptions'

PR #116, opened 2018-07-19
Change Name to ID

mentzerk:
*Issue #, if available:*
Not available

*Description of changes:*
A ref to a security group returns the group ID, not the group name.  This example was broken as-is.  Changing to Id allows the example to execute.

### DIFF
--- a/doc_source/aws-properties-ec2-security-group-ingress.md
+++ b/doc_source/aws-properties-ec2-security-group-ingress.md
@@ -164,7 +164,7 @@ The following template snippet creates an EC2 security group with an ingress rul
                 "IpProtocol": "tcp",
                 "FromPort": "80",
                 "ToPort": "80",
-                "SourceSecurityGroupName": {
+                "SourceSecurityGroupId": {
                     "Ref": "SGBase"
                 }
             }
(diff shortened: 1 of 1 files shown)


rjlohan:
Per the EC2 documentation, this Ref is actually a conditional return value (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html#w2ab2c21c10d475c15);

> When you specify an AWS::EC2::SecurityGroup type as an argument to the Ref function, AWS CloudFormation returns the security group name or the security group ID (for EC2-VPC security groups that are not in a default VPC).

I can't see for sure that your change is more or less correct than the existing example, given that context matters.

mentzerk:
@cmmeyer - Yes - I can rebase.
@rjlohan - I'm interested in the conditional return, as it would seem that's what this was originally written against, but when I used the example code - the cloudformation threw an error.  Perhaps theres a bigger bug somewhere and the logic on the conditional return has broken?  I can double check the example and see if it's still happening before moving forward.

mentzerk:
Going to close this PR, my suggested change was made in this commit: https://github.com/awsdocs/aws-cloudformation-user-guide/commit/640230f21f1d1d2813057db4dacb1f1d5c980317

PR #117, opened 2018-07-24
Security Group example syntax updates

Alex-Burgess:
*Issue #, if available:*
Really minor, but will possibly save people from cleaning syntax to prevent errors/warnings when using IDEs that have syntax checking, e.g. Atom with atom-cfn-lint.

The examples show FromPort and ToPort as strings, when the specification has them as integers.  When I copied the example I had to clean the syntax to avoid errors.

*Description of changes:*
Changed FromPort and ToPort to integer values instead of strings in the examples section. 


ckesser:
Thanks for the contribution! Merging your request.

PR #123, opened 2018-07-30
the mixing of Arn and ARN is driving me crazy.

edyesed:
please accept this PR as a  feature request for every API/configuration item which refers to ARN as Arn.

* AMI rhymes with Cherry Pie
* ARN is pronounced Aye-Arrr-Ennn

*Issue #, if available:*

The [reciprocal PR is here](https://github.com/awsdocs/aws-cloudformation-user-guide/pull/124), in case you are one of those people who feel that the approach contained herein in backwards. 

*Description of changes:*

- The Mixing of ARN with Arn is driving me bananas, particularly inside the AWS::Config:: namespace, where my `AWS::Config::ConfigurationRecorder` wants it to be spelled **RoleARN**, and my `AWS::Config::ConfigurationAggregator` wants it to be spelled **RoleArn**

I'm unwilling to die upon this hill, and I realize that there may be APIs that need to change behind the scenes, but this is madness. 


ckesser:
@edyesed Thanks for voicing this concern; we understand your frustration. Unfortunately, this inconsistency stems from the source code, and the docs have to reflect the service API exactly (including case sensitivity). I'm going to close out these requests. Again, we so appreciate you identifying this. Keep the feedback coming!

PR #132, opened 2018-08-09
fix ec2 launchtemplate to match specifications of the api

nappa32:
*Issue #, if available:*

*Description of changes:*
The AWS::EC2::LaunchTemplate resource property has wrong specification:
 - Max Length: Actual 125  vs. 128 in docs
 - Allowed characters : [a-zA-Z0-9\(\)\.-/_]+ VS. [a-zA-Z0-9\(\)\./_]+

### DIFF
--- a/doc_source/aws-resource-ec2-launchtemplate.md
+++ b/doc_source/aws-resource-ec2-launchtemplate.md
@@ -37,8 +37,8 @@ Properties:
 
 `LaunchTemplateName`  <a name="cfn-ec2-launchtemplate-launchtemplatename"></a>
 A name for the launch template\.  
-Length Constraints: Minimum length of 3\. Maximum length of 128\.  
-Pattern: \[a\-zA\-Z0\-9\\\(\\\)\\\.\-/\_\]\+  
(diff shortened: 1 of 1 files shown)


rjlohan:
This property's constraints are taken directly from https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateLaunchTemplate.html

We'll have to confirm with the service team if this is accurate.

nappa32:
Has anything come from this PR ?

nappa32:
Please note the error that I'm experiencing is : 

```
InvalidLaunchTemplateName.MalformedException | The specified launch template name is invalid. A launch template name                                     must be between 3 and 125 characters, and may contain letters, numbers,                                     and the following characters: ()./_.
```
This is a difference as to what is stated in the docs. The link I supplied is from AWS ec2 docs as well. https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html#api-error-codes-table-client ( search for : InvalidLaunchTemplateName.MalformedException ).

PR #134, opened 2018-08-09
Remove (erroneous) syntax spec in overview

edenhochbaum:
Syntax description here isn't really necessary, as it's just an overview.  And in this case, the guidance is specific to JSON definitions, while CloudFormation supports at least one other format (YAML.)


ckesser:
@edenhochbaum Thanks for this! I completely agree. Merging your request now.

PR #135, opened 2018-08-09
Remove incorrect assertion about parameters

edenhochbaum:
You can define resources conditionally without a parameters section.

### DIFF
--- a/doc_source/conditions-section-structure.md
+++ b/doc_source/conditions-section-structure.md
@@ -13,10 +13,7 @@ During a stack update, you cannot update conditions by themselves\. You can upda
 
 ## How to Use Conditions Overview<a name="w3ab2c17c15c21c13"></a>
 
-To conditionally create resources, you must include statements in at least three different sections of a template:
-
-`Parameters` section  
-Define the input values that you want to evaluate in your conditions\. Conditions will result in true or false based on values from these input parameters\.
+To conditionally create resources, you must include statements in at least two different sections of a template:
 
 `Conditions` section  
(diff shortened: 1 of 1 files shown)


rjlohan:
Hi @edenhochbaum, thanks for this submission. While you're technically correct in this change, the docs are setup to outline what we determined to be the most useful path for Conditions. Namely; using input parameters to control runtime execution of a template. Without using Parameters, what use case would be satisfied by Conditions?

edenhochbaum:
> Without using  Parameters, what use case would be satisfied by Conditions?

Typically, you'd be creating a resource conditioned on the value of a pseudo-parameter, such as AWS::Region.  For example, if I want to create a VPC in us-east-1 and no other regions, I wouldn't use a Parameters section.

Separately, I think updating technical documentation to make it "technically correct" is worth doing.  Just for pragmatic reasons, I think it's counter-productive to make a false statement in technical docs for pedagogical reasons.

rjlohan:
That's a fair use-case, though I'd rather call this out more explicitly in the docs rather than hide this further. e.g; extend your changes to include some text around including template or pseudo-parameters. That would improve the docs beyond the technical accuracy.

ckesser:
I'd like to merge this request with some modifications based on the above conversation:

"Depending on the entity you want to conditionally create or configure, you must include statements in the following template sections:
Parameters section
Define the inputs that you want your conditions to evaluate. The conditions evaluate to true or false based on the values of these input parameters. If you want your conditions to evaluate pseudo parameters, you don't need to define the pseudo parameters in this section; pseudo parameters are predefined by AWS CloudFormation."

Thanks for helping us clarify @edenhochbaum!

PR #136, opened 2018-08-10
Add example bucket metrics for entire S3 bucket

adamcousins:
*Issue #, if available:*

*Description of changes:*
Add example bucket metrics for entire S3 bucket


ckesser:
Thanks for this contribution @adamcousins! I'd like to merge your request with some modifications; we can slip a 'MetricsConfigurations' specification into an existing example rather than including it as a separate example, as I've done in the 'Configure a Static Website with a Routing Rule' example here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket.html#aws-resource-s3-bucket-examples. Thanks again, and keep the contributions coming!

PR #143, opened 2018-08-23
Mark ServiceRole as mandatory

alexjurkiewicz:
If you attempt to update an existing `AWS::ElasticBeanstalk::Application` resource without a `ResourceLifecycleConfig` property defined, adding one without `ServiceRole` defined will result in the error `Parameter ServiceRole is invalid. Must be a valid IAM Role ARN.` (and rollback will fail with the same error). There might be some corner cases where this parameter isn't required, but it seems easiest to mark it as mandatory (especially when 99% of use-cases will use the service-linked account that's automatically created).


ckesser:
@alexjurkiewicz Thanks for your feedback! I reached out to our product team and they confirmed that 'ServiceRole' is conditionally required if 'VersionLifecycleConfig' is also specified. I'm going to merge this request and make a couple modifications to elaborate on the conditions under which the property is required. Thanks again!

alexjurkiewicz:
@ckesser thank you for merging this. Did you ever get to make the extra changes you talked about?

PR #144, opened 2018-08-28
Update aws-properties-lambda-function-code.md Fix YAML example typo

lennynyktyk:
Hello,

Not sure if

```yaml
key: >
  foo
  bar
```

is valid syntax.  I think the `>` should be a `|`.


rjlohan:
Thanks for the submission, however the existing content is correct as it stands. 

Both styles are valid YAML, per 1.1 the spec for scalar content here; http://yaml.org/spec/1.1/#id858081

The current syntax more accurately reflects the corresponding JSON examples; the JavaScript code is concatenated by spaces, because the JavaScript includes statement-terminating characters (;) and so can be wrapped to a single line. However in the subsequent Python example, the strings are concatenated with newlines, because in Python whitespace matters.

Both `>` and `|` are valid characters but imply a particular type of wrapping, so context matters.

PR #145, opened 2018-08-30
Be clear about what happens when no revision is specified

patrobinson:
*Description of changes:*

If we don't supply a revision the most recent revision is selected. Further updates to the task definition are therefore do not take effect on the Event Rule.


alexjurkiewicz:
I wonder if this will make people think CloudFormation will automatically update the resource if you update the task definition by hand and then update the stack. A more explicit formulation might be "If no Task revision is supplied it defaults to the most recent revision *at time of resource creation.*"

rachfop:
Hello, thank you for your suggestions. The page you referenced has been updated to reflect your suggestion. You see those changes [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-events-rule-ecsparameters.html)

PR #155, opened 2018-09-10
Correct valid json in KMS Key

seittema:
Minor fix: json syntax block has the comma in the wrong spot for the properties


ckesser:
Thanks for catching this, @seittema! Merging your request now.

PR #157, opened 2018-09-11
Removing line numbers from json/yaml examples

bundyfx:
*Description of changes:*
There are inconsistencies when it comes to JSON/YAML examples. Some of the examples include line numbers which make it tedious to copy and paste. In order to streamline the usability of our examples, I propose that we remove line numbers from code example areas. Also, where possible the structure of our JSON examples could be run through a "beautifier" which format's into a standardized structure. 


ckesser:
Thanks for these fixes, @bundyfx! We appreciate you removing the line numbers for ease of copy/paste. We also appreciate your feedback about standardizing our example structure. We'll definitely keep an eye out for JSON structure that could be improved for readability. Merging your request.

PR #161, opened 2018-09-22
Fix SpotFleet LaunchSpecifications / LaunchTemplateConfigs requirement

calebmarcus:
Previously, LaunchSpecifications was listed as required -- but LaunchTemplateConfigs can now be used instead. I tried this with CF, and it complained when both were specified. So, LaunchSpecifications shouldn't be listed as "Required: Yes"


ckesser:
Thanks for updating those property requirements, @calebmarcus! I'll merge this.

PR #163, opened 2018-09-24
SpotFleet LaunchTemplateConfig requires LaunchTemplateSpecification

calebmarcus:
LaunchTemplateSpecification is listed as not required in LaunchTemplateConfig, but when omitted, CF returns: "At least one LaunchTemplateConfig does not have LaunchTemplateSpecification." (This makes sense -- a LaunchTemplateConfig wouldn't be much good without a launch template to refer to.)


ckesser:
Thanks for this one too @calebmarcus, merging.

PR #167, opened 2018-09-26
Update aws-resource-logs-destination.md

alexjurkiewicz:
Use native YAML for IAM policy, and explain why the name is mandatory.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/doc_source/aws-resource-logs-destination.md
+++ b/doc_source/aws-resource-logs-destination.md
@@ -108,13 +108,6 @@ DestinationWithName:
     DestinationName: "TestDestination"
     RoleArn: "arn:aws:iam::123456789012:role/LogKinesisRole"
     TargetArn: "arn:aws:kinesis:us-east-1:123456789012:stream/TestStream"
-    DestinationPolicy:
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
I reverted "Use native YAML for IAM policy" because `DestinationPolicy` requires a string, not object. Unfortunate :(.

I think it would be good to more explicitly call this out in its docs, as it doesn't match the approach used by most of the rest of CloudFormation, where IAM policies are supported as native objects.

alexjurkiewicz:
ping

ckesser:
I'm definitely going to merge your explanation of the DestinationName requirement. Thanks for adding some valuable context! However, I'm leaning toward keeping the DestinationPolicy description as-is; we explicitly state which data type each property accepts, so I think additionally stating which data types a property _doesn't_ accept might read as redundant rather than clarifying. If you have thoughts about this, we can always revisit in the comments here or in a new PR.

alexjurkiewicz:
That makes sense to me. Thanks for your time today!! 🙇

PR #181, opened 2018-10-26
Add Pattern restriction docs for IdentityPoolName

christemple:
*Issue #, if available:*

I couldn't figure out why my IdentityPoolName was failing validation with - in it as from the docs there only appeared to be a length restriction.

*Description of changes:*

I finally found the pattern restriction on the property [here](https://docs.aws.amazon.com/cognitoidentity/latest/APIReference/API_CreateIdentityPool.html#API_CreateIdentityPool_RequestSyntax), and have updated the cloudformation docc to include the Pattern.


ckesser:
Thanks for adding this, @christemple! Merging your request now.

PR #190, opened 2018-11-09
add ACM certificate timeout warning

glb:
[Setting AWS CloudFormation Stack Options](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-console-add-tags.html) says:

> By default, there is no timeout for stack creation. However, individual resources may have their own timeouts based on the nature of the service they implement. For example, if an individual resource in your stack times out, stack creation also times out even if the timeout you specified for stack creation has not yet been reached.

It would be very helpful to have the resource timeout (currently appears to be 12 hours) included in the documentation for this resource.

*Issue #, if available:*

*Description of changes:*
Add a note about the resource timeout for this resource.

### DIFF
--- a/doc_source/aws-resource-certificatemanager-certificate.md
+++ b/doc_source/aws-resource-certificatemanager-certificate.md
@@ -5,6 +5,8 @@ The `AWS::CertificateManager::Certificate` resource requests an AWS Certificate 
 **Important**  
 When you use the `AWS::CertificateManager::Certificate` resource in an AWS CloudFormation stack, the stack will remain in the `CREATE_IN_PROGRESS` state and any further stack operations will be delayed until you validate the certificate request, either by acting upon the instructions in the certificate validation email, or by adding a CNAME record to your DNS configuration\.
 
+If the certificate request is not validated within XXX hours, the stack will time out and be rolled back, deleting the pending certificate.
+
 **Topics**
 + [Syntax](#aws-resource-certificatemanager-certificate-syntax)
 + [Properties](#aws-resource-certificatemanager-certificate-properties)


ckesser:
Thanks for working with us to improve our docs! Although timeout specifics would be useful, especially for those resources with a lengthy timeout duration, resource timeouts are subject to change over time due to a variety of environmental factors. This means we're unable to pinpoint precise, consistent timeouts to incorporate into the docs without over-complicating things. I'm going to close out this PR for the time being. Thank you again for the excellent feedback :)

glb:
@ckesser thanks for the comment! Would you accept a modified update that says that there *is* a resource timeout for this resource? As it stands the documentation led me to believe that the stack would wait indefinitely for the certificate to be approved -- which seemed reasonable given that multiple teams may need to collaborate globally for approvals to happen. I was unpleasantly surprised when attempting to coordinate a certificate request with my IT team and the resource timeout happened while our internal ticket was being processed.

ckesser:
@glb Ah, I see now how that "Important" note is misleading. Thanks for getting back to me and pushing for clarity here! If I understand the root of your confusion correctly, I'm thinking a simple addition like the bold text below might clear things up:

Important
When you use the AWS::CertificateManager::Certificate resource in an AWS CloudFormation stack, the stack will remain in the CREATE_IN_PROGRESS state and any further stack operations will be delayed until you validate the certificate request, either by acting upon the instructions in the certificate validation email, or by adding a CNAME record to your DNS configuration, **or until resource creation times out**.

Let me know if you have any other ideas!

glb:
@ckesser that will help, thank you!

PR #202, opened 2018-11-24
Remove incorrect quotation mark

cariad:
The YAML example block incorrectly contained a single quotation mark. I have removed it.


ckesser:
Thanks for the fix, @cariad!

PR #203, opened 2018-11-27
Clarify wording of domain name requirements for API Gateway resource

Kausheel:
*Description of changes:*

Users have indicated unexpected stack failures when using capital letters in the DomainName property. The stack error message is not clear enough to indicate that capital letters are not allowed. A use case for why capital letters may be inserted is if the user is trying to Fn::Sub the stack name into the DomainName property, and it's common for stack names to have capital letters. Adding this change would prevent confusion for future users.

### DIFF
--- a/doc_source/aws-resource-apigateway-domainname.md
+++ b/doc_source/aws-resource-apigateway-domainname.md
@@ -50,7 +50,7 @@ The reference to an AWS\-managed certificate for use by the edge\-optimized endp
 *Update requires*: [No interruption](using-cfn-updating-stacks-update-behaviors.md#update-no-interrupt)
 
 `DomainName`  <a name="cfn-apigateway-domainname-domainname"></a>
-The custom domain name for your API in Amazon API Gateway\.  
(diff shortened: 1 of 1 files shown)


doc_source/aws-resource-apigateway-domainname.md
```diff
@@ -50,7 +50,7 @@ The reference to an AWS\-managed certificate for use by the edge\-optimized endp
 *Update requires*: [No interruption](using-cfn-updating-stacks-update-behaviors.md#update-no-interrupt)
 
 `DomainName`  <a name="cfn-apigateway-domainname-domainname"></a>
-The custom domain name for your API in Amazon API Gateway\.  
+The custom domain name for your API in Amazon API Gateway\.The domain must be lower case\.
```
chandanadesilva: perhaps, 'The domain name must be lowercase'

ckesser:
Thanks for the note, @Kausheel. Merging your contribution.

PR #206, opened 2018-11-29
Listener Rule Condition Values Clarification

taylorb-syd:
*Issue #, if available:* N/A

*Description of changes:* Updated the wording on the listernerrules-condition properties page to better clarify what the values are. This is because it may be confusing for some customers who try and combine multiple `host-headers` together that they need to refer to the API to know how many values are accepted by a given `Field`.


ckesser:
Hey @taylorb-syd, I'm going to merge this with modifications (I'll do you one better by bringing over the complete property descriptions from the ELB API Reference: https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_RuleCondition.html). Thanks for helping us flesh out these descriptions!

PR #193, opened 2018-11-13
added T3 instances for options related to CPU credits

tyron:
*Issue #, if available:*

*Description of changes:*
Adding T3 instances where we talk about credit options for EC2 T2 family.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #207, opened 2018-11-29
added T3 instances for options related to CPU credits

tyron:
*Issue #, if available:*

*Description of changes:*
Adding T3 instances where we talk about credit options for EC2 T2 family.
(this is to replace #193)


ckesser:
Supersedes #193

PR #212, opened 2018-12-07
fixed wrong description of DynamoDB LSI index name

ytakahashi:
*Description of changes:*

IndexName of DynamoDB LSI can't contain symbols other than `-`, `_`, `.`. so I fixed the description.
(I tried and I got an error saying `Member must satisfy regular expression pattern: [a-zA-Z0-9_.-]+ `)


ckesser:
@ytakahashi Good catch, thanks for the update! Merging now.

PR #214, opened 2018-12-12
Move EC2-VPC to the main function description

danielhodder:
EC2-VPC is the default for all new accounts and has been for many years. This function, just reading the description and examples, is misleading in that case since it doesn't return the availability zones in the region, but rather those that have default subnets. This change seeks to make this more obvious since that's the default case, with EC2-Classic being essentially a legacy deployment mechanism these days


rachfop:
Thank you for your submission. This PR addressed a needed piece of information.

PR #215, opened 2018-12-13
Complete list of Type values for ELB Listener

pbudzon:
*Issue #, if available:* N/A

*Description of changes:* `Type` property only listed one valid value. Completed the list to include all possible values.


ckesser:
We so appreciate your contribution, @pbudzon! This definitely needs to be updated; I'll merge your request now. Thanks!

PR #217, opened 2018-12-18
Mark the FromPort and ToPort options to a Security Group rule as required for TCP, UDP, and ICMP

danielhodder:
*Description of changes:*

Mark the FromPort and ToPort options to a Security Group rule as required. Not specifying these results in this errors from the CloudFormation engine:

> Invalid value 'Must specify both from and to ports with TCP/UDP.' for portRange

> Invalid value 'Must specify both from and to ports with ICMP.' for portRange


ckesser:
Thanks for your contribution, @danielhodder! I'm going to verify this with our service team to make sure the docs and the [resource spec](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) stay in sync. I'll leave an update here shortly.

ckesser:
Hi @danielhodder, 

Thank you so much for your patience. We updated our doc source to better reflect the requirements. Feel free to either rebase (so I can merge this contribution) or close the PR.

danielhodder:
Seems like the documentation has been updated on master with this information that is in this PR.

PR #218, opened 2018-12-20
Make it more clear which property to change

jenseickmeyer:
When creating a LaunchTemplate via CloudFormation for the first time I set a list of Security Group IDs with the parameter SecurityGroups since it stated

> For a nondefault VPC, you must use security group IDs instead.

For me, it was not obvious that the parameter SecurityGroupIds was meant.


ckesser:
This definitely improves clarity. Thanks for contributing, @jenseickmeyer! I'll merge your request now. Keep the feedback coming!

PR #219, opened 2018-12-24
fix(node): invalid and not supported node version 4.3

rastvaho:
Node version for this template leads to an error. You need to use higher version. This template needs to be updated.


ckesser:
Thanks for the update, @rastvaho! I'll merge your contribution.

PR #220, opened 2018-12-28
Escaping slash in BinaryMediaTypes list

bushong1:
*Issue #, if available:*

*Description of changes:*
Adding a clarifying statement to the docs of BinaryMediaTypes to ensure developers know they MUST escape all `/` with `~1`.  Adding Parameter Defaults to the example that includes BinaryMediaTypes.


bushong1:
This has been addressed in the documentation

bushong1:
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings-configure-with-control-service-api.html

PR #227, opened 2019-01-04
Remove obsolete whitespace

fatbasstard:
Remove an obsolete whitespace in the JSON example. Details, but annoying if you would copy/paste this as a starting point (since you'll get an error when deploying this. Except when using a linter for example, like [cfn-python-lint](https://github.com/awslabs/cfn-python-lint))

(the other EOL whitespaces are removed by my editor setting, but since I think that makes sense., I kept that in the PR)

image


ckesser:
Great catch, greater meme! Thank you for this contribution! Merging now.

PR #228, opened 2019-01-06
State and ExecutionRoleArn are required

fabiodouek:
*Issue #, if available:*

*Description of changes:*
State and ExecutionRoleArn are required. Minor change.

### DIFF
--- a/doc_source/aws-resource-dlm-lifecyclepolicy.md
+++ b/doc_source/aws-resource-dlm-lifecyclepolicy.md
@@ -42,7 +42,7 @@ The description of the lifecycle policy\.
 
 `ExecutionRoleArn`  <a name="cfn-dlm-lifecyclepolicy-executionrolearn"></a>
 The Amazon Resource Name \(ARN\) of the IAM role used to run the operations specified by the lifecycle policy\.   
- *Required*: No  
+ *Required*: Yes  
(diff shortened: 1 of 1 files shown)


cmmeyer:
Hi @fabiodouek -- on what are you basing this change?  The CloudFormation resource specification and the underlying DLM API (https://docs.aws.amazon.com/dlm/latest/APIReference/API_LifecyclePolicy.html) also have these properties marked as `Required: No`.  Are you getting an error that contradicts this?

fabiodouek:
Hi @cmmeyer , It seems that the link you've referenced is the data type.

I'm based on the CreateLifecyclePolicy API: https://docs.aws.amazon.com/dlm/latest/APIReference/API_CreateLifecyclePolicy.html , which mentions Role State and ExecutionRoleArn being required.

What prompted me to review the documentation, was because when I created the CloudFormation stack without defining those two parameters, I got an exception during the stack creation, complaining that those attributes were not defined.

I hope that this makes sense.

cmmeyer:
Thanks for the clarification. It looks like we'll need to update both the docs and the spec for this one then.

ckesser:
@fabiodouek, thanks for the fix! @cmmeyer, thanks for the approval! Merging.

PR #229, opened 2019-01-07
Clarify when types are a list.

sturdiva:
*Documentation does not make it clear that certain properties require a list of objects, not a single object.*


ckesser:
Thanks for correcting this inconsistency, @sturdiva! Merging your PR.

PR #230, opened 2019-01-07
Update aws-resource-ec2-ec2fleet.md to correct expected types of Vali…

Canuteson:
Correct expected typs of ValidFrom and ValidUntil properties of EC2Fleet objects from integer to string

*Description of changes:*
The Documentation describes these as ISO formatted UTC datetime strings but incorrectly states the property must be integers. CloudFormation rejects integers for these properties, and expects and ISO8601 formatted string, similar to `SpotFleetRequestConfigData`


ckesser:
Thanks for this update, @Canuteson! We really appreciate the contribution. Merging your PR.

PR #233, opened 2019-01-08
Update cfn-whatis-howdoesitwork.md

ryanliu1021:
Add an available file extension type that is missing in the paragraph, but it's supported by Cloudformation.


rjlohan:
This is fine, though CloudFormation doesn't actually care what file extension you use - this documentation is just for example.

PR #245, opened 2019-01-11
Add Valid Values information to the Route53 ResolverEndpoint direction

fatbasstard:
Based on the API Reference (As mentioned on top of the page) added the `Valid value` information (just like other property documentation) of the Route53 Resolver endpoint's direction

Docs:
* https://docs.aws.amazon.com/Route53/latest/APIReference/API_route53resolver_ResolverEndpoint.html
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-route53resolver-resolverendpoint.html#cfn-route53resolver-resolverendpoint-direction

(Fyi: My editor removes EOL whitespaces)


ckesser:
Thanks for another excellent contribution!

PR #246, opened 2019-01-11
Improve S3 WebsiteConfiguration documentation

fatbasstard:
The `IndexDocument` is not mandatory. See the Screenshot from a Stack deployment:

image

Updated the documentation to reflect this:

* Made the Required of the IndexDocument `Conditional`
* Added description that IndexDocument, ErrorDocument and RoutingRules are not allowed to be specified if `RedirectAllRequestsTo` is specified (already mentioned at the `RedirectAllRequestsTo`, but this makes it a bit more explicit


fatbasstard:
@ckesser @rabilahiri 

Just reconstructed the PR since it was outdated

fatbasstard:
PR keeps getting outdated because the entire docs gets recreated. I'll close the PR since keeping it "mergeable" is an ongoing thing

PR #247, opened 2019-01-14
set type 'Schedule' for schedules

robpickerill:
*Issue #, if available:*
N/A - minor change

*Description of changes:*
The example Cloudformation stack uses a schedule of type `Default`. This results in:

```
'schedule' values must be of type 'Schedule'. Found values of type 'Default'
```

Changing this to be of type `Schedule` resolves the issue - https://docs.aws.amazon.com/datapipeline/latest/DeveloperGuide/dp-object-schedule.html


ckesser:
Thanks for this fix, @robpickerill! I'll merge your contribution.

PR #256, opened 2019-01-21
adding a note about explicit tagging

ebs-mkhludnev:
Updating the doc in according to the info provided us in AWS Support ticket. ie ```
Unfortunately, it is not possible to inherit all the Tags set to the Stack automatically. The only possible way is to provide the "Tags" property under the "ComputeResources" property.```

*Issue #, if available:*

*Description of changes:*


rachfop:
Resolving, I believe the docs address your pull request now. If not, feel free to make a pull request:

Key\-value pair tags to be applied to EC2 resources that are launched in the compute environment\. For AWS Batch, these take the form of "String1": "String2", where String1 is the tag key and String2 is the tag value−for example, \{ "Name": "AWS Batch Instance \- C4OnDemand" \}\. This is helpful for recognizing your AWS Batch instances in the Amazon EC2 console\. These tags can't be updated or removed after the compute environment has been created; any changes require creating a new compute environment and removing the old compute environment\. These tags are not seen when using the AWS Batch `ListTagsForResource` API operation\.

PR #263, opened 2019-02-01
Update aws-properties-ecs-taskdefinition-tmpfs.md

wdegeus:
Change required to Yes for tmpfs size, since without the Size parameter it will fail. Note that the API (https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Tmpfs.html) has required Yes for size as well.

*Issue #, if available:*

*Description of changes:*


ckesser:
Hi @wdegeus, 

Thank you for your contribution! Thank you for your contribution! The "Type", "Required", "and "Update behavior" fields for each property are now auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) will be updated after the specification is updated. Feel free to keep in touch here.

PR #266, opened 2019-02-05
Private APIs in APIG don't support custom domains

azec-pdx:
According to the [Private API Development Considerations](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-private-apis.html?shortFooter=true#apigateway-private-api-design-considerations), the private API Gateway doesn't support custom domain names:
> Custom domain names are not supported for private APIs.

*Issue #, if available:*

*Description of changes:* Removing mention of the custom domains in relation to private API endpoint.


ckesser:
Yep, thanks for this contribution! Merging now.

PR #268, opened 2019-02-06
Added region to table

zepplen:
*Issue #, if available:*

*Description of changes:*
Added region to CF specification table to allow people scraping the page to associate the cf specification directly to the region.
Changed the name of the old region column to Region Name which is inline w/ the labeling on the RDS region's page.


ckesser:
Hey @zepplen, I really like this suggestion. I'm going to merge your contribution with some additional changes; I'll probably reconfigure the table a bit while I'm at it.

PR #278, opened 2019-02-22
Update aws-resource-rds-dbcluster.md

foolproofit:
Add information to MasterUserPassword property that changes are not applied immediately but in the upcoming maintenance window.

*Issue #, if available:*
CASE 5811100811

*Description of changes:*
For Aurora, when modifying a DB cluster, only changes to the DB cluster identifier, IAM DB authentication, and New master password settings are affected by the Apply Immediately setting [1].
Currently CloudFormation does not have property for DB cluster by which  'Apply Immediately' flag can be set and hence the password gets updated during upcoming maintenance window (not immediately).

Resources:
[1] https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.Modifying.html

### DIFF
--- a/doc_source/aws-resource-rds-dbcluster.md
+++ b/doc_source/aws-resource-rds-dbcluster.md
@@ -189,7 +189,7 @@ The master user name for the DB instance\.
 The password for the master database user\.  
 *Required*: Conditional\. You must specify this property unless you specify the `SnapshotIdentifier` property\. In that case, do not specify this property\.  
 *Type*: String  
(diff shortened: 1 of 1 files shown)


randyurbano:
Your change is for CreateDBInstance, which doesn't have the ApplyImmediately parameter. ModifyDBInstance has the ApplyImmediately parameter. When a user creates a DB instance, the specified password is set for the DB instance during DB instance creation. Can I close the pull request (without merging)? Let me know if you have questions or comments.

PR #279, opened 2019-02-27
Casing is wrong which causes the property to fail

charliejllewellyn:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/doc_source/aws-properties-ecs-service-networkconfiguration.md
+++ b/doc_source/aws-properties-ecs-service-networkconfiguration.md
@@ -8,14 +8,14 @@
 
 ```
 {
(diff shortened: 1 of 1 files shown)


ckesser:
Hi @charliejllewellyn, 

Thanks so much for the feedback. I'm looking at the **NetworkConfiguration** property definition in the public CFN resource spec, and it looks like the casing in the docs is correct. The **AwsvpcConfiguration** property, lowercase 'v', of **AWS::ECS::Service.NetworkConfiguration** takes an **AwsVpcConfiguration** object, uppercase 'V'. This is the capitalization used in the [ECS docs](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_AwsVpcConfiguration.html), as well.

PR #281, opened 2019-02-28
Indicate that StackSets limits are soft limits

ikben:
This adds the wording used for indicating that the number of stacks can be increased to stack sets and stack instances.

IIRC there is no form for this yet in the "Limit Increase" drop downs, but a normal support case should work.


ckesser:
Hi @ikben,

Will you please rebase so I can merge your contribution? Thanks for adding this!

ikben:
@ckesser No problem, done!

PR #283, opened 2019-03-04
Update aws-properties-ec2-instance-launchtemplatespecification.md

pranitr:
I have tested this with CFN that if I give Launch Template Name for LaunchTemplateId, I get an error saying, I must give Launch Template ID for the LaunchTemplateId. 
And same for the LaunchTemplateId, It needs Launch Template ID and not the Launch Template Name.
Previously the docs was suggesting that, you can give either Launch Template Name or Launch Template ID for both the properties, hence corrected the same.

*Issue #, if available:*

*Description of changes:*


ckesser:
Hi @pranitr, 

Thank you for your feedback. You must specify either a template ID or a template name, meaning that either the **LaunchTemplateId** or the **LaunchTemplateName** property must be specified. I see how the note might read as valid values for each property instead, and I'd be happy to make this text clearer. I'm going to merge your contribution with a few of my own changes. Thanks again.

PR #284, opened 2019-03-05
change 'name' property of Ulimit to be required

Brijeshrpatel9:
Description of changes: 'Name' property in Ulimit section for task definition is required property. When I did not pass 'Name', cloudformation gave me an error. I found that it's required parameter from API reference page: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Ulimit.html


ckesser:
Taking a look at the resource spec, this is accurate. Thanks for fixing this, @Brijeshrpatel9! I'll merge your contribution.

PR #288, opened 2019-03-07
Update example snippet

svdgraaf:
*Description of changes:*
Updated the example snippet, in the current form it doesn't work, the `statusCode` should be an integer. With the current template, this will be the error:

`Execution failed due to configuration error: statusCode should be an integer which defined in request template`

Also, that error sentence is missing a word (`which [is] defined`), might want to have that checked as well...


ckesser:
Hi @svdgraaf,

Thank you for the fix! I'll merge your contribution. I'll also look into that error message.

PR #290, opened 2019-03-09
Clarifies Target Group values for HealthCheckTimeoutSeconds and UnhealthyThresholdCount

lorengordon:
*Description of changes:*

Spent an inordinate amount of time troubleshooting why my Target Group kept failing to create with a cryptic message like this:

> Internal failure (Service: AmazonElasticLoadBalancingV2; Status Code: 500; Error Code: InternalFailure; Request ID: <request-id>)

Tracked it down to differences between TCP/TLS and HTTP/HTTPS protocols in the Target Group and the allowed values in the health check. Figured I'd try to clarify the docs the docs to hopefully reduce the time someone else spends troubleshooting this problem.


julieso:
Thanks for the suggestions. The original equated each type of load balancer with a set of supported target group protocols. This version is clearer and more complete.

lorengordon:
Thanks!

PR #294, opened 2019-03-15
Correct AWS::SSM::PatchBaseline.RejectedPatchesAction to String

fatbasstard:
The `RejectedPatchesAction` property of the `AWS::SSM::PatchBaseline` is a String, not an Array.

Sources:
- AWS Spec files (Huge files so no link)
- API: https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_CreatePatchBaseline.html#systemsmanager-CreatePatchBaseline-request-RejectedPatchesAction
- Console:
```
Property validation failure: [Value of property {/RejectedPatchesAction} does not match type {String}]
```


ckesser:
Accurate! Thanks for the fix. I'll merge.

PR #303, opened 2019-03-27
Fixes inconsistances of capitalization for Ec2TagSet

ajoga:
*Issue #, if available:*

*Description of changes:*
The Ec2TagSet property is inconsistently capitalized in the docs. It lead me to open case ID 5919886811 because of confusion. Please note that I think EC2TagSetListObject may also be victim of the same confusion, but I haven't tested it. 
Also, the capitalization may need to be fixed in the API reference, but I'm unsure : https://docs.aws.amazon.com/codedeploy/latest/APIReference/API_EC2TagSet.html 


ckesser:
Thanks, @ajoga! This corrects the docs to match the [resource spec](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). Merging your contribution.

PR #304, opened 2019-03-28
Fix the example in `AWS::EC2::NatGateway`'s document

aoyagi-faber:
*Issue #, if available:* None

*Description of changes:*
The `AWS::EC2::EIP` resource document
(https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-eip.html)
says that

> If you define an Elastic IP address and associate it with a VPC that is
> defined in the same template, you must declare a dependency on the
> VPC-gateway attachment by using the DependsOn attribute on this resource.

And the `DependsOn` attribute document
(https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-dependson.html)
also says that Elastic IP addresses depends on a VPC-gateway attachment.

This pull request fixes the example in `AWS::EC2::NatGateway`'s document to
follow the above constraint.

### DIFF
--- a/doc_source/aws-resource-ec2-natgateway.md
+++ b/doc_source/aws-resource-ec2-natgateway.md
@@ -75,7 +75,6 @@ The following example creates a NAT gateway and a route that associates the NAT 
 
 ```
 "NAT" : {
-  "DependsOn" : "VPCGatewayAttach",
   "Type" : "AWS::EC2::NatGateway",
   "Properties" : {
     "AllocationId" : { "Fn::GetAtt" : ["EIP", "AllocationId"]},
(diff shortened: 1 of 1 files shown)


ckesser:
Thanks for the correction, @aoyagi-faber! Would you mind rebasing so I can merge your contribution?

aoyagi-faber:
@ckesser, what I intended to correct seems to be already contained in a commit in master.
Please check the output of `git log -1 -b -p master^{/"06 May 2019"} doc_source/aws-resource-ec2-natgateway.md`, and close this pull request after you make sure that this is true.

PR #306, opened 2019-04-03
Modifying VPC flow log docs to reflect what is currently true due to …

dominic-pucci:
…return values from Log Group resource

*Description of changes:*

When using the `LogDestination` property and supplying the ARN of a CloudWatch Log Group via the !GetAtt function, the return value is an invalid ARN of the format `arn:aws:logs:us-east-1:123456789012:log-group:/mystack-testgroup-12ABC1AB12A1:*`

This was confirmed by AWS support to cause the Access Denied errors that I was getting when trying to add VPC flow logs via CloudFormation. Using the `LogGroupName` property instead fixed the issue, and these updates are to ensure others do not hit the same issue as I did.

### DIFF
--- a/doc_source/aws-resource-ec2-flowlog.md
+++ b/doc_source/aws-resource-ec2-flowlog.md
@@ -53,8 +53,8 @@ If you specify `LogDestinationType` as `s3`, do not specify `DeliverLogsPermissi
 *Update requires*: [Replacement](using-cfn-updating-stacks-update-behaviors.md#update-replacement)
 
 `LogDestination`  <a name="cfn-ec2-flowlog-logdestination"></a>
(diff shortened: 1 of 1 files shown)


katerini:
Hello, thanks for submitting this PR! 

In this case, you can still use the `LogDestination` property and specify the ARN for a log group, but as you rightly noted, the ARN must not include `:*` in the syntax. When I tested, it accepted the ARN in this form:
`arn:aws:logs:us-east-1:123456789012:log-group:my-logs`

Perhaps we should rather update the description for `LogDestination` to use an example of what would work, and suggest LogGroupName as an alternative? For example:

"If LogDestinationType is not specified or `cloud-watch-logs`, specify the Amazon Resource Name (ARN) of the CloudWatch Logs log group. For example, to publish to a log group called `my-logs`, specify `arn:aws:logs:us-east-1:123456789012:log-group:my-logs`. Alternatively, omit `LogDestination` and specify `LogGroupName` instead."

Hopefully that should help prevent further issues like yours.

PR #307, opened 2019-04-07
ServiceRole requires an Arn

gergnz:
ServiceRole requires an Arn, and while Ref to a Role works it's technically incorrect. It doesn't work if the path is NOT '/'.

*Issue #, if available:*

*Description of changes:*

Use FnGetAtt(GetAtt) and not Ref in the last example to retrieve the ARN.


ckesser:
Thanks, @gergnz! Merging your contribution.

PR #311, opened 2019-04-18
Fix Targets Key=InstanceIds

wckzhang:
The Key value for Targets cannot be instanceids, instead, it must be InstanceIds.


rachfop:
Hello, thank you for your feedback. You can see your changes [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-maintenancewindowtask.html).

PR #314, opened 2019-04-25
Update for using Custom Roles

badaldavda8:
I tested that now we can use custom roles with StackSets on both Administrator account and Target Accounts. I see that we have added a note that they must be named certain names, but this is no longer true if we use custom roles.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/doc_source/stacksets-prereqs.md
+++ b/doc_source/stacksets-prereqs.md
@@ -15,11 +15,14 @@ Because stack sets perform stack operations across multiple accounts, before you
 
 1. Create the necessary IAM service roles in your administrator and target accounts to define the permissions you want\.
 **Important**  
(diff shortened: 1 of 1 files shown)


rachfop:
This was a great suggestion and we apologize for not getting to it sooner. The Topics section, followed by the two topics is a convention of the CloudFormation documentation. The documents have received considerable updates, including: The role in your administrator account _should_ be named **AWSCloudFormationStackSetAdministrationRole**. The role in each of your target accounts _should_ be named **AWSCloudFormationStackSetExecutionRole**. [source](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/stacksets-prereqs-self-managed.html).

Thank you for your suggestions.

PR #316, opened 2019-04-25
Update eks resourcesvpcconfig properties

hub714:
Any update to resourcesvpcconfig requires replacement.

*Issue #, if available:*

*Description of changes:*
Changed no interruption to replacement for eks resourcesvpcconfig properties. There isn't currently an update mechanism for eks.


ckesser:
Hi @hub714, 

Thank you for your contribution! The "Type", "Required", and "Update behavior" fields for each property are now auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) will be updated after the spec is updated. Feel free to keep in touch here.

PR #317, opened 2019-04-26
fixed code snippet for custom resource section, example3.yaml

andrew-ha:
Changed the example 3 yaml section for the file "aws-resource-cfn-customresource.md" from this:
```
    ServiceToken:
      !Sub |
        arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${LambdaFunctionName}
```
To this:
```
    ServiceToken:
      !Sub arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${LambdaFunctionName}
```

The character "|" is adding a "\n" to the end of the service token, which is causing stacks to fail.

### DIFF
--- a/doc_source/aws-resource-cfn-customresource.md
+++ b/doc_source/aws-resource-cfn-customresource.md
@@ -153,8 +153,7 @@ MyCustomResource:
   Type: "Custom::TestLambdaCrossStackRef"
   Properties: 
     ServiceToken:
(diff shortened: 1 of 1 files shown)


rjlohan:
Thanks, good spot. `|` is a YAML token to create multi-line strings - which isn't valid here, so this is the right fix.

PR #318, opened 2019-04-26
BillingMode property applies to base table

edenhochbaum:
This is just to clarify that the BillingMode property isn't a property of the GSI, but rather of the GSI's base [parent?](https://github.com/awsdocs/amazon-dynamodb-developer-guide/issues/55) table.


rachfop:
I resolved conflicts on this pull request. Thank you for your suggestion.

PR #326, opened 2019-05-06
[kms] Fix maximum PendingWindowInDays

michael-k:
*Issue #, if available:* n/a

*Description of changes:* In the console and all other resources, the maximum `PendingWindowInDays` for KMS Keys is 30, not 365.

----


juneb:
Hi, Michael. Unfortunately, we can't accept the PR. If we make the change in the markdown docs here, it will be overwritten the next time the source docs are generated and converted to markdown. 

This discrepancy appears because the object type of the PendingWindowInDays parameter takes values from 1-365, but KMS code separately enforces a limit of 7-30. To fix it, we need to change the object type or use a new object type. I've filed an issue with the KMS engineering to fix this. 

Thanks again for your help.

PR #353, opened 2019-05-24
removing AWS/Amazon prefixes

PatMyron:
*Issue #, if available:*

*Description of changes:*
Reducing inconsistencies/redundancies in service names by removing AWS/Amazon prefixes


ckesser:
Hi @PatMyron, 

Unfortunately, we can't ditch the AWS/Amazon prefixes in some cases; it's a legal requirement. For readability, the resource reference contains as few prefixes as we can get away with, and we alphabetize by the "short" name of the service even when the prefix is present.

PR #358, opened 2019-05-27
fix: Predicate.Type indentation

clifflu:
`Predicate.Type` should have same indentation as other predicate attributes, namely, `DataId` and `Negated`.


rachfop:
Hello, thank you for your feedback. You can see your changes [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-waf-rule.html).

PR #359, opened 2019-05-28
Document interruption for DBInstanceClass

72636c:
*Issue #, if available:* N/A

*Description of changes:* I don't know the specifics here, but there is definitely some interruption during a change in instance class.


ckesser:
Hi @72636c, 

Thank you for your contribution! The "Type", "Required", "and "Update behavior" fields for each property are now auto-generated based on the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) will be updated after the specification is updated. Feel free to keep in touch here.

PR #360, opened 2019-05-28
Fix EC2 NetworkInterface AssociatePublicIpAddress

adamchainz:
*Description of changes:*

The description was some copypasta from ipv6 addresses. Updated to a description adapted from the same property on launch templates: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-networkinterface.html


julieso:
Thanks for pointing this out.

PR #385, opened 2019-06-21
Update aws-resource-ecs-service.md

ZainubAmod:
Updated documentation to avoid customer confusion. If Task definition revision is not specified, service will create correctly BUT CFN ECS service resource will NOT stabilize, causing stack creation to fail.

*Issue #, if available:*

*Description of changes:*


ZainubAmod:
Change is pushed through to docs

PR #387, opened 2019-06-25
Clarified AWS::ApiGateway::Deployment resource behaviour re StageName/StageDescription.

taylorb-syd:
*Issue #, if available:* N/A

*Description of changes:*

It is not possible to perform an update to the `StageName` property at all, and often when trying to change the `StageDescription` properties you will get the following error.

```
StageDescription cannot be specified when stage referenced by StageName already exists
```

This is because the `StageDescription` and `StageName` properties for a Deployment are all about creating a new stage. We should discourage the usage of Stages in Deployments and encourage consumers to use `AWS::ApiGateway::Stage`.


taylorb-syd:
Honestly I'm of the opinion we should remove the ability to create Stages in Deployments all-together, however this should clear up some confusion using the behaviour.

PR #395, opened 2019-07-04
Fixed wrong Property name

steamrolly:
*Issue #, if available:* N/A

*Description of changes:*
Fixed wrong Property name.


rachfop:
Hello, thank you for your suggestion. Every bit of feedback helps us put our focus where it is needed the most. You can see your request was updated [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-wafregional-ratebasedrule.html).

PR #397, opened 2019-07-07
Reword best practices on organization

adamchainz:
*Issue #, if available:* n/a

*Description of changes:* It's not "normal" if it's the best practice not to, more "initial"


ckesser:
Good call. Thank you! Merging.

PR #400, opened 2019-07-10
Fix the type of Essential

viraptor:
*Description of changes:*

Property `Essential` is documented as a Boolean type at:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions.html#cfn-ecs-taskdefinition-containerdefinition-essential
Examples used string instead.


joelbrandenburg:
Thanks for the documentation contribution!

PR #403, opened 2019-07-15
Proposed change to make it more clear that the policy is an object

timmattison:
This PR illustrates an area where the CloudFormation documentation could use some work to make it more clear when adding inline policies. The documentation references "Json" in both the JSON and YAML areas which has led some of my customers to believe that this value needs to be a long JSON string.

This shows up in the IAM documentation and likely other places as well.

Instead of referencing Json we should call out that the value is actually an object in whichever syntax the customer is using. This would help if we later add an additional syntax as well.

### DIFF
--- a/doc_source/aws-resource-iot-policy.md
+++ b/doc_source/aws-resource-iot-policy.md
@@ -12,7 +12,7 @@ To declare this entity in your AWS CloudFormation template, use the following sy
 {
   "Type" : "AWS::IoT::Policy",
   "Properties" : {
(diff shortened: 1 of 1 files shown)


rachfop:
Hello, thank you for your feedback. I will take note of it and pass it along to the CloudFormation team. Currently, The "Type", "Required", and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) -- so we are unable to change the type property Json to Object.

PR #404, opened 2019-07-18
Quote the map keys in the example so that they are parsed as strings and not as numbers

vissree:
*Issue #, if available:*

*Description of changes:*
The map keys under the TimeBasedAutoScaling property in the given example are not quoted and hence will be parsed as numbers. If we follow this example in an actual template, the CloudFormation run fails with the below template validation error.
Template format error: [/Resources/Instance/Properties/TimeBasedAutoScaling/Monday] map keys must be strings; received numeric [1] instead

This commit quotes the map keys in the given example.


GitGaby:
Thanks, Vishnu; looks good.

PR #408, opened 2019-07-24
Correction for AllocatedStorage in CF template

bebzzz:
Correction for AllocatedStorage and DBInstanceClass. 
Seems that "AllocatedStorage" can't be 5 now. It was minimum for MagneticStorage
Now SSD GP2 is default with minimum 20Gb storage size
DBInstanceClass=db.m1.small  doesn't exist now
---
RDS does not support creating a DB instance with the following combination: DBInstanceClass=db.m1.small, Engine=mysql, EngineVersion=5.7.22, LicenseModel=general-public-license. For supported combinations of instance class and database engine version, see the documentation. (Service: AmazonRDS; Status Code: 400; Error Code: InvalidParameterCombination; Request ID:

*Issue #, if available:*

*Description of changes:*


rachfop:
Hello @bebzzz our templates now include `db.t2.small` as `db.m1.small` has deprecated and `db.m3` is a previous generation standard instance class.

PR #417, opened 2019-07-31
Fix invalid reference to security group

tomzo:
It seems most of the page was updated, but this example is invalid and fails because GroupId: requires an Id to be extracted from the object first.


julieso:
Thanks for the contribution.

PR #419, opened 2019-08-05
removing trailing whitespace from doc_source/

PatMyron:
This should produce more readable diffs and fewer merge conflicts, especially with pull requests originating from Github's browser editor 

https://stackoverflow.com/questions/149057/how-to-remove-trailing-whitespace-of-all-files-recursively

1776/1776 files in `doc_source` modified with whitespace-only changes:

Screen Shot 2019-08-05 at 10 43 51 AM


PatMyron:
Discussed this change internally and the [next periodic update](https://github.com/search?q=user:awsdocs+"periodic+update"&type=Commits) could re-introduce the trailing whitespace. Investigating the work needed to strip trailing whitespace during the build process

PR #422, opened 2019-08-09
AWS::ElasticLoadBalancingV2::TargetGroup -- Protocol is required

alanivey:
For `AWS::ElasticLoadBalancingV2::TargetGroup`, protocol is required; otherwise CloudFormation returns:

```
A protocol must be specified (Service: AmazonElasticLoadBalancingV2; Status Code: 400; Error Code: ValidationError
```

*Description of changes:*

Simple change:

```diff
- *Required*: No
+ *Required*: Yes
```


ckesser:

PR #423, opened 2019-08-12
AWS::RDS::DBCluster MasterUserPassword conditionally required

PatMyron:
`Property MasterUserPassword cannot be empty.`

[`AWS::RDS::DBCluster`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html)


ckesser:
Because this value is automated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html), which reflects that MasterUserPassword is optional, I've submitted a ticket to retrieve a comprehensive list of conditions under which MasterUserPassword is required. Closing this in favor of the ticket.

PR #434, opened 2019-08-20
Update aws-resource-rds-dbsubnet-group.md

rhucho:
While trying to launch a CloudFormation stack, I got the following error :

    "DB subnet group 'DB Subnet name' does not exist. (Service: AmazonRDS; Status Code: 404; Error Code: DBSubnetGroupNotFoundFault; Request ID: e82696ed-50be-432d-b8a8-6bc7082e7a44)"

On further testing, I found that this issue occurs if the value of the "DBSubnetGroupName" property has Upper case characters (Ex - In my template : "DB Subnet name"). Even though the value of the "DBSubnetGroupName" property has Upper case characters (Ex : "DB Subnet name"), RDS creates a subnet group with all the characters stored as a lowercase string [1] (Ex : "db subnet name"). However, CloudFormation hasn't converted the characters from upper case to lower case. So, when CloudFormation tries to create a DB Cluster and tries to attach the DBSubnetGroup named "DB Subnet name", the DescribeDBSubnetGroups API call fails with this error : "DB subnet group 'DB Subnet name' does not exist". The reason being the underlying DB Subnet Group's name has all lowercase characters, however, CloudFormation calls the DescribeDBSubnetGroups API with the value of the DBSubnetGroup property - as it was specified in the template (Doesn't convert it to a lowercase string).

Hence, the documentation needs to be changed to ensure that the "DBSubnetGroupName" property doesn't contain any Upper case characters.

REFERENCE:
~~~~~~~~~~~~~~~~~
[1] DBSubnetGroupName :
    https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbsubnet-group.html#cfn-rds-dbsubnetgroup-dbsubnetgroupname

*Issue #, if available:*

*Description of changes:*



rachfop:
Thank you for the suggestion. The document has been improved.

PR #436, opened 2019-08-24
Update aws-resource-wafregional-ratebasedrule.md

ferarevalo:
The Yaml example is incorrect, 'Predicates' should be 'MatchPredicates' as showed in the Resource definition and the JSON example.


judypear:
Thank you for this fix!

PR #463, opened 2019-09-15
AWS::ServiceDiscovery::Service NamespaceId required

PatMyron:
```
Service namespace id cannot be null (Service: AWSServiceDiscovery; Status Code: 400; Error Code: InvalidInput; Request ID: REDACTED)
```
```yaml
Resources:
  Resource:
    Type: AWS::ServiceDiscovery::Service
```


ckesser:

PR #472, opened 2019-09-17
Remove integer bounds from string properties

cariad:
*Description of changes:*

Removed the integer "minimum" and "maximum" bounds from string properties.


chrisdr:
These numbers refer to the minimum and maximum allowed length of the string you set as the value of the parameter. Although these length limits look strange for parameters like "treatmissingdata" which has a limited set of valid values, they are still in the code as enforced limits, so I'm going to leave them in the document as is.

Thanks for using AWS, and thanks for taking time to help us make our docs better.

cariad:
I appreciate the feedback, thank you!

PR #483, opened 2019-09-25
Denoting AuthorizationType as Required as per SDK

scottbrown:
When you try to create an APIGateway::Method resource without an authorizationType set, the underlying SDK will fail with error:

```json
...
"ResourceStatusReason": "1 validation error detected: Value null at 'putMethodInput.authorizationType' failed to satisfy constraint: Member must not be null (Service: AmazonApiGateway; Status Code: 400; Error Code: ValidationException; Request ID: b0c6d8f2-0554-4b21-9f02-ea42d28e78b4)",
...
```

yet the documentation here says that it is not required.  This change updates that mistake so that the field is required.


ckesser:
The "Type", "Required", "and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). We have a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-method.html#cfn-apigateway-method-authorizationtype) will be updated after the specification is updated. Thank you for this feedback!

Issue #490, opened 2019-10-02
Include Supported Regions for PropertyTypes and ResourceTypes


ScriptAutomate:
## Issue

AWS documentation seems to be out-of-sync across a variety of places when it comes to:

  - What AWS services are supported in each region
  - What AWS services are supported by AWS CloudFormation in each region

Instead, it exists across a spectrum of sources that can contradict each other at times:

  - [AWS Region Table](https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services/)
  - [AWS Service Pricing](https://aws.amazon.com/pricing/)
  - [AWS CFN Resource Specification Files](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html)
  - [AWS Announcements](https://aws.amazon.com/new) (ex. _"...the following regions are supported for this **new** service..."_)
  - [AWS Documentation](https://docs.aws.amazon.com/) (ex. _"...the following regions are supported for this service..."_)

## Proposed Solution

For AWS CloudFormation, the source-of-truth seems to be the [AWS CloudFormation Specification Files](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) (also seen here, but the "source" hasn't been updated to reflect production since March 2019: [markdown source](https://github.com/awsdocs/aws-cloudformation-user-guide/blob/master/doc_source/cfn-resource-specification.md) and currenlty referenced in issues like #426 ).

I have an auditing repository, [AWS CloudFormation Resource Specification Auditor](https://github.com/ScriptAutomate/aws-cfn-resource-specs), that has generated a JSON containing what resources / properties are supported by what regions: [ScriptAutomate/aws-cfn-resource-specs/blob/master/supported-regions-per-resource.json](https://github.com/ScriptAutomate/aws-cfn-resource-specs/blob/master/supported-regions-per-resource.json)

If the CFN docs workflow has something similar to what I'm doing with the proof-of-concept tool, [aws-doc-region-updater.py](https://github.com/ScriptAutomate/aws-cfn-resource-specs/blob/master/tools/aws-doc-region-updater.py) (could be split into functions and cleaned up), the docs can automatically contain the latest supported region information at each `ResourceSpecificationVersion` release.

PatMyron:
[...]

PR #491, opened 2019-10-02
Adding Supported Regions to All Docs Based on CFN Spec File Inclusion

ScriptAutomate:
*Issue #, if available:*

#490 

*Description of changes:*

I have run a tool to add `## Supported Regions` as a new header to each PropertyType and ResourceType doc, including whether all regions are supported or a bullet-point list of known supported regions. Supported regions are generated from auditing all [CFN Resource Spec files](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html).


PatMyron:
@ScriptAutomate

1) I'm not sure the CloudFormation Resource Specification currently aims to track regional availability down to the property type level. I believe property types are included in a region's resource specification if their respective resource type is included in the region's resource specification. 
_Property type level information should probably be omitted from the documentation while this is still the case._

2) While it is aimed for, CloudFormation Resource Specifications are not always accurate with regards to regional availability at the resource type level. As one example, it is possible to create some resource types in a region even while they are not present in that region's resource specification.

Those caveats aside, **I absolutely love this idea**! This information would be incredibly useful in resource type documentation and provide even more motivation to maintain the accuracy of the resource specifications with respect to regional availability.

The [Documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html), [CDK](https://github.com/aws/aws-cdk), [GoFormation](https://github.com/awslabs/goformation), and [Troposphere](https://github.com/cloudtools/troposphere) currently rely on the `us-east-1` resource specification, but the [CloudFormation Linter](https://github.com/aws-cloudformation/cfn-python-lint) relies on every region's resource specification and would also benefit greatly from that work.

ScriptAutomate:
Thanks for the feedback, @PatMyron

> 1. I'm not sure the CloudFormation Resource Specification currently aims to track regional availability down to the property type level. I believe property types are included in a region's resource specification if their respective resource type is included in the region's resource specification.
>    _Property type level information should probably be omitted from the documentation while this is still the case._

Got it. I could modify the tool to not include property types when it comes to adding supported regions.

> 2. While it is aimed for, CloudFormation Resource Specifications are not always accurate with regards to regional availability at the resource type level. As one example, it is possible to create some resource types in a region even while they are not present in that region's resource specification.
> Those caveats aside, **I absolutely love this idea**! This information would be incredibly useful in resource type documentation and provide even more motivation to maintain the accuracy of the resource specifications with respect to regional availability.

I see. This is interesting. I'm wondering if this would be the preferred path:

- Use the CFN resource spec files as a current 'master' even though it is known that it isn't always perfect, which fits with the philosophy behind having the documentation source on GitHub
- Add `## Supported Regions` markdown header/anchor to each resource type doc (leaving property types alone), and include the regions that list the resource in their spec files. Default to saying _all regions are supported_ if the resource is found across all spec files
- With the public docs reflecting the CFN resource spec files, any docs that are found by users as not listing specific regions that they know are supported (by either opening an issue or PR to append a missing region or removing one) could perhaps double as issues/PRs for adding resources to the CFN spec files. This fits well with where you say: _"This information would be incredibly useful in resource type documentation and provide even more motivation to maintain the accuracy of the resource specifications with respect to regional availability."_
  - I've seen this being done when it comes to other PRs, such as corrections to params that were originally listed as required, and are no longer default-required. The documentation PRs end up getting closed with a note that the CFN spec files will be fixed to reflect the proper param information, due to being used as the source of info in the AWS CFN docs ( an example of this outcome can be seen in #464 )
  - This approach would be something of a best-guess documentation that can have issues/PRs raised, where incorrect, by the community. Documentation can be auto-remedied when CFN spec files have resources added if something akin to [ScriptAutomate/aws-cfn-resource-specs/aws-doc-region-updater.py](https://github.com/ScriptAutomate/aws-cfn-resource-specs/blob/master/tools/aws-doc-region-updater.py) was included as part of an AWS internal pipeline for building/pushing the documentation.

Thoughts? I could close this PR, and I could run the tool against the latest `6.3.0` release with just resource types targeted, and ref this PR and the accompanied issue? I could just close the PR and continue the conversation in the GitHub issue, being that this PR has acted more as a POC anyway.

ScriptAutomate:
I'm going to close this for now, due to current documentation conflicts and the need to remove propertytype regional availability in the first pass. I will reference this PR (and the related issue) in the future with modified PR(s) that are helpfully generated by the continued development of [ScriptAutomate/aws-cfn-resource-specs/aws-doc-region-updater.py](https://github.com/ScriptAutomate/aws-cfn-resource-specs/blob/master/tools/aws-doc-region-updater.py)

PR #493, opened 2019-10-05
Fix formatting error In YAML version of example

Ro5635:
*Issue #, if available:*

The tags 'Value' attribute should have the same indentation as the 'Key' to which it relates, see the JSON version of the example immediately above in the doc to see that they should both be in the same array nested within the tags array. :)

*Description of changes:*

Added the necessary indentation to make it part of the correct tag, and **not** a top level resource in the Cloud Formation template


Ro5635:
After looking more carefully through the long long list of open PR's on this repo I have found this is a duplicate  :sweat_smile: . So I'm closing this.

PR #496, opened 2019-10-11
documenting CloudFormation template minification strategy

PatMyron:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html


rachfop:
Thank you for your suggestion. Minification is a strategy that can help with reducing the size of a template.

PR #497, opened 2019-10-12
AWS::Route53Resolver::ResolverRule RuleType always FORWARD

PatMyron:
This value is always FORWARD. Other resolver rule types aren't supported.


skralik:
Thanks for the heads up. I updated both the Route 53 API Ref and the CloudFormation doc about Resolver to note that only Resolver can create a rule that has a RuleType of RECURSIVE. I had already updated the doc with information about SYSTEM rules.

skralik:
The updated doc will be published to GitHub next week.

PR #500, opened 2019-10-15
Fix inner link

gurdiga:
*Issue #, if available:*

One of the _internal_ links in the text is broken (probably outdated).

*Description of changes:*

Since it’s an internal link, I’ve cut it down to the anchor name: from `https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#cfn-iam-role-templateexamples` to `#aws-resource-iam-role--examples`, which seems less code to maintain; I hope that’s alright.

---


rachfop:
Hello, thank you for your suggestion. The page has been updated and the links now work. You can see the page [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html).

PR #509, opened 2019-10-30
Clarification of requirements needed to get it to work

Senneseph:
*Issue #, if available:*
Incomplete documentation.

*Description of changes:*
Added: Note indicating how to get the things specified in the AWS::CloudFormation::Init property to be applied to the EC2 instance (for Linux).

I spent 5+ hours going in circles with the existing documentation.  It was extremely confusing that something called "init" was not actually being initialized by default.


PatMyron:
Agreed this is not intuitive. [Another part of the documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-init.html#cfn-init-Examples) touches on this, but it would be a good addition to this page as well because there is clearly [still confusion around this](https://stackoverflow.com/questions/14454657/awscloudformationinit-how-does-it-work)

rachfop:
Thank you for clarifying this.

PR #514, opened 2019-11-07
Fix node require() path for cfn-response

djamison:
*Issue #, if available:*

*Description of changes:*
The current code `require('cfn-response')` produces a `Cannot find module` error. A relative path needs to be used because Cloudformation places the `cfn-response.js` file at the root level with the code from the `ZipFile` parameter.

The bug will leave your Cloudformation stack in a broken state because the custom resource cannot be successfully created (causing a rollback) nor deleted (leaving the stack in a rollback cleanup state until timed out, etc.).

image


mwunderl:
The service bug that caused this error has been resolved. Please let us know if you still see any issues.

Thanks,
-Michael at AWS

PR #517, opened 2019-11-13
Update: Allowed OAuth flows instuctions

mattbryce93:
The instructions for setting the AllowedOAuth flows are incorrect, rather than setting this value to token or code. In actuality, the value must be, according to the error below an array of strings with only the values of "implicit", "client_credentials", "code".

`1 validation error detected: Value '[token]' at 'allowedOAuthFlows' failed to satisfy constraint: Member must satisfy constraint: [Member must satisfy enum value set: [implicit, client_credentials, code]] (Service: AWSCognitoIdentityProvider; Status Code: 400; Error Code: InvalidParameterException;`

```
AllowedOAuthFlows

    Set to code to initiate a code grant flow, which provides an authorization code as the response. This code can be exchanged for access tokens with the token endpoint.

    Set to token to specify that the client should get the access token (and, optionally, ID token, based on scopes) directly.

    Required: No

    Type: List of String

    Maximum: 3

    The update requires: No interruption
```

*Issue #, if available:*

*Description of changes:*


julieso:
Thanks for the contribution!

PR #518, opened 2019-11-13
Fix: AllowedOAuthScopes parameter documentation

mattbryce93:
`Invalid scope requested: Cognito (Service: AWSCognitoIdentityProvider; Status Code: 400; Error Code: ScopeDoesNotExistException;`

This should be possible based on the docs stating:
```
AllowedOAuthScopes
A list of allowed OAuth scopes. Currently supported values are "phone", "email", "openid", and "Cognito". In addition to these values, custom scopes created in Resource Servers are also supported.

Required: No

Type: List of String

Maximum: 50

Update requires: No interruption
```

The actual scopes available for creation with CloudFormation (not including custom scopes) are actually:

aws.cognito.signin.user.admin
phone
openid
email
profile


julieso:
Thanks for the contribution. I added more detail from the API documentation.

PR #520, opened 2019-11-15
Update gettingstarted.templatebasics.md

kennyk65:
YML format doesn't require all of those extra quotes.  There are a few cases where the quotes are needed, but not the casual use shown here.

*Issue #, if available:*

*Description of changes:* See comment

### DIFF
--- a/doc_source/gettingstarted.templatebasics.md
+++ b/doc_source/gettingstarted.templatebasics.md
@@ -162,20 +162,20 @@ One of the greatest benefits of templates and AWS CloudFormation is the ability 
 ```
 Resources:
   Ec2Instance:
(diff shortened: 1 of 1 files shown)


rachfop:
Thank you very much for your feedback @kennyk65 . We appreciate your pull and we take into consideration format changes like this. 
I am going to merge this pr, but the AWS CloudFormation designer does put ' ' on strings by default. So those extra ' ' around strings in YAML may still show up at a later time. The integers can be removed. 

We are looking at a way to systematically update our current system to reflect your changes. I am going to close [your other pr ](https://github.com/awsdocs/aws-cloudformation-user-guide/pull/16) which address the ' ', and pass this information on.
Thank you again for your feedback.

Issue #534, opened 2019-12-01
Deployment type should be required in Documentation in AWS::CodeDeploy::DeploymentGroup DeploymentStyle


keroloswilliam:
As mentioned in this [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codedeploy-deploymentgroup-deploymentstyle.html#cfn-codedeploy-deploymentgroup-deploymentstyle-deploymenttype) defining DeploymentOption and DeploymentType Is Not Required 
But After Defining DeploymentType to be IN_PLACE for 2 instances I got the Error 
```
Deployment type or deployment option cannot be null or empty 
``` 
So DeploymentType can not be Null make sense since i have to choose the type and there is no default mentioned in any place 

but DeploymentOption should accept null since i can deploy without being behind load balance Also if you are working with console you won't find that option when choosing IN_PLACE as Deployment Type 

I have already reached the support to open internal bug and this Issue just to clarify

ckesser:
Closing per #536

QasimArthuna53:
I had the same problem, and I couldn't able to solve it.
this is my codedeploy target

`DeploymentStyle: 
        DeploymentType: IN_PLACE
        DeploymentOption: Yes`

but this throws this error 

`Deployment option can only be WITH_TRAFFIC_CONTROL or WITHOUT_TRAFFIC_CONTROL`

Help will be appreciated.

Thank you

PR #536, opened 2019-12-01
CHANGE the required option to Yes for DeploymentType and DeploymentOptions

keroloswilliam:
Signed-off-by: kerolos <kerolos.william@gmail.com>

*Issue #534 

*Description of changes:*
Change the required option for DeploymentOption and DeploymentType to be Yes since both are required and cloudformation don't set a default


ckesser:
The "Type", "Required", "and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codedeploy-deploymentgroup-deploymentstyle.html) will be updated after the specification is updated. Feel free to keep in touch here.

Ochaun:
This is still an issue...

PR #537, opened 2019-12-03
Update URLs in cfn-helper-scripts-reference.md to S3 virtual host style

rtkjbillo:
*Issue #, if available:* No issue

*Description of changes:* As per https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html, use virtual-style S3 URLs rather than path-based URLs to access the CloudFormation example resources.

Since these buckets appear to be located in us-east-1, I also validated that there was no 307 redirect when accessing the individual objects with `curl`.


rachfop:
Thank you for the suggestion. The docs are now up to date.

phils:
@rachfop Did you merge this in reverse?  The changes were reverted in your commit before merging, so the fix to the docs never went in and we just got whitespace changes.  Thanks.

PR #544, opened 2019-12-13
Change Example to use DomainName instead of DomainID

badaldavda:
***Issue #, if available:*** 

Using this example from 234-246 for creating record fails with error message -
"    "errorMessage": "[Tried to create an alias that targets e1fsf78lgyjqau., type A in zone Z2FDTNDATAQYW2, but the alias target name does not lie within the target zone]",
"

***Description of changes:***
Example in this document does a Ref on CloudFront Distribution - 

DNSName:
          Ref: myCloudFrontDistributionDomainName

whereas as per - 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudfront-distribution.html#aws-resource-cloudfront-distribution-return-values

Ref gives a CloudFront distribution id - E27LVI50CSW06W instead of DNS record. 

It should be GetAtt with Attribute DomainName which gives Domain Name

DNSName:
          Fn::GetAtt: myCloudFrontDistribution.DomainName

Attaching the files that worked and the one that failed.

I hope this helps.


badaldavda:
[Archive.zip](https://github.com/awsdocs/aws-cloudformation-user-guide/files/3960336/Archive.zip)

Files that work and files that done

rachfop:
Hello, thank you for your submission. I can approve of the pr as soon as your resolve the conflicts and rebase. Thank you.

PR #552, opened 2019-12-18
Update required flag for MetricName

janosmeszaros:
*Issue #, if available:*
None

*Description of changes:*
Fixing inconsistency in CloudWatch documentation. 


ckesser:
Hi @janosmeszaros! Thanks for the feedback. I'm working with my service team to update the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html), from which we automate the property requirements in the docs. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) will be updated after the resource specification is updated. Feel free to keep in touch here.

PR #554, opened 2019-12-28
Made TriggerName required

Patrick-56Bit:
Received an error from CloudFormation that TriggerName was not set when I didn't included it. Everything worked well when it was included. Kindly confirm with the CloudFormation product team and accept change to docs if this change is accurate.

*Issue #, if available:*

*Description of changes:*


ckesser:
The "Type", "Required", "and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codedeploy-deploymentgroup-triggerconfig.html#cfn-codedeploy-deploymentgroup-triggerconfig-triggername) will be updated after the specification is updated. Feel free to keep in touch here.

PR #556, opened 2019-12-31
Clarify that AZ can't be update for the EBS Volume

uthark:
Clarify that AvailabilityZone can't be update for the EBS Volume

If AvailabilityZone is updated, CloudFormation returns the following error:

```
UPDATE_FAILED                 Volume properties other than AutoEnableIO, type, size, and IOPS cannot be updated. Change the properties back to previous values and update the stack again.
``` 


julieso:
Thanks for the contribution. The only properties that can be updated are the volume size, type, and IOPS.

PR #557, opened 2020-01-03
Update CloudFormation docs for ECS container dependencies on Fargate

itwasntandy:
*Issue*
Currently the CloudFormation documents for [Container Dependency](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdependency.html) seems to be outdated, as it states that Container Dependencies are not available on Fargate.

The [ECS docs for Container Dependencies](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_dependson) state that it is available as of Fargate `1.3.0`.

*Description of changes:*

Updating CloudFormation  docs to reflect ECS, which I presume to be source of truth.

Additionally updating that, I noticed an inconsistency for the [CF Container Definitions docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions.html#cfn-ecs-taskdefinition-containerdefinition-dependson) too, where what is in GitHub doesn't match what's publicly displayed, so fixing that too.


joelbrandenburg:
Thanks for your contribution and for using Amazon ECS!

PR #572, opened 2020-01-24
Update aws-properties-sns-topic.md

lantrix:
Seems like this should be SNS subscriptions.

*Description of changes:*

 - Typo fixed


seittema:
@lantrix  Actually the documentation is correct.  DisplayName is an attribute of an SNS topic for the name you see when dealing with SMS Subscriptions.  The TopicName is the name of the SNS topic itself.

devlinbd2:
Thank you for flagging the potential issue. It looks like it could be a typo, but it is actually correct as written. You can configure an Amazon SNS topic to distribute as SMS messages to mobile devices.  As pointed out by @lantrix, the DisplayName will be displayed to the device.

PR #574, opened 2020-01-27
Fixed contradiction in API Gateway CloudFormation docs

jameslamb:
*Description of changes:*

The [API Gateway Stage docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-stage.html) currently have conflicting information about the `DeploymentId` parameter. The description contains the text `This parameter is required.` and 'Required: No`. I have seen, from CloudFormation errors, that the parameter is required...in this PR I propose changing the documentation to reflect that.

I'm sorry that additional lines were touched by this PR (the `Resources:` examples). Looks like  it may be related to line endings.

### DIFF
--- a/doc_source/aws-resource-apigateway-stage.md
+++ b/doc_source/aws-resource-apigateway-stage.md
@@ -89,8 +89,8 @@ The ID of the client certificate that API Gateway uses to call your integration 
 *Update requires*: [No interruption](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-no-interrupt)
 
 `DeploymentId`  <a name="cfn-apigateway-stage-deploymentid"></a>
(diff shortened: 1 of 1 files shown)


ckesser:
The "Type", "Required", "and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). There's a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-stage.html#cfn-apigateway-stage-deploymentid) will be updated after the specification is updated.

jameslamb:
Ok thanks. @ckesser is there a better and more direct way for me to report these issues in the future?

ckesser:
@jameslamb Not that I'm aware of. @PatMyron ?

PR #580, opened 2020-02-06
updated the correct service name

sumanth-lingappa:
The correct service is com.amazonaws.{region}.s3 as there will be a S3 presigned URL will be generated.

*Issue #, if available:*

*Description of changes:*


seittema:
@sumanth-lingappa The documentation is actually correct.  The page is for the cloudformation interface endpoints which are in fact:  com.amazonaws.{region}.cloudformation

PR #584, opened 2020-02-12
`RepositoryUrl` must not cointain an username

ppihus:
By default, Bitbucket, and likely others, give the HTTPS URL for a repository together with the username. Such as `https://myusername@bitbucket.org/mycompany/myrepository.git`

While `AWS::SageMaker::CodeRepository` seems accepts this format, `AWS::SageMaker::NotebookInstance` fails when referencing the added `CodeRepository` in ` DefaultCodeRepository` (probably the same applies when referencing it in `AdditionalCodeRepositories`).

As the Secret used by `AWS::SageMaker::CodeRepository` has `username` defined in it anyway, I removed the username from the Git repository URL and this way `AWS::SageMaker::NotebookInstance` did not fail anymore when attaching the git repository.

Interestingly, failure reasons were different yesterday from what they are today. However, removing the username from Git repository URL has solved the problems for now.

*Issue #, if available:*

*Description of changes:*
Sidenote in the documentation that RepositoryUrl must not contain the username.


aws-rbs:
Thanks Pritt. Updating the docs with the note.

PR #588, opened 2020-02-17
Fixing wafv2 ipset

seittema:
When working with the WafV2 ipset I discovered the following issues.  These were all tested through trial and error with a sample cft stack.

* Type for the Addresses was incorrect, it's a list of string
* Missing the - on the yaml examples for list
* Type for the Tags was incorrect, its a typical list of Tag
* Properties pages for ipset no longer needed after fixing the typing


ckesser:
Thank you, @seittema!

PR #589, opened 2020-02-17
wafv2 Rule Name required

seittema:
If you do not include a Name for the rule, rather than auto-generate, CFT returns the following error: 
"1 validation error detected: Value null at 'rules.1.member.name' failed to satisfy constraint: Member must not be null (Service: Wafv2, Status Code: 400,"

Therefore Name should be marked as required


ckesser:
Hi @seittema! Thank you for the feedback. The source of this information, the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html), has been updated with your change. The documentation will automatically update shortly.

PR #600, opened 2020-02-24
Remove references to TagList in AWS::WAFv2

lewayne-aws:
Remove TagList on AWS::WAFv2::WebACL

Remove TagList on AWS::WAFv2::RegexPatternSet

Remove TagList on AWS::WAFv2::IPSet

Remove TagList on AWS::WAFv2::RuleGroup

*Issue #, if available:*

*Description of changes:*
Throughout the current documentation for AWS::WAFv2, there are references to a TagList type. This is inconsistent with the rest of the documentation, and is misleading. Just like with other resources that can be tagged, this should be a list of tags.

Another issue that is fixed with this is that the current documentation's use of TagList makes it appear that "TagList" can be placed into a template, which is not the case.


ckesser:
These changes were merged with #588. Thank you for these fixes, @lewayne-aws!

lewayne-aws:
I looked over that commit, and it does not cover all of the instances of TagList. I have rebased my fork to line up with master so it can be reviewed again.

PR #605, opened 2020-02-25
Add documentation for PrivateIPv4Address on SubnetMapping

lewayne-aws:
AWS::ElasticLoadBalancingV2::LoadBalancer supports a PrivateIPv4Address
for Network Load balancers. This is mentioned in
aws-resource-elasticloadbalancingv2-loadbalancer.md but is not actually
documented anywhere.


julieso:
Thanks for pointing this out. The documentation is generated from the resource specification. The documentation will be updated after the resource specification is updated.

PR #609, opened 2020-02-29
Fix example SAML template: email capitalization

ozydingo:
Just confirmed on my project that the `email` attribute mapping for a SAML IdP should be lowercase (like every other example), not uppercase. Using `Email` causes the stack for fail: "missing require attribute `email`"


julieso:
Thanks for your contribution!

PR #611, opened 2020-03-03
Fix typo in example: !Ref

perlpunk:
I think this is a typo:
```
Value: Ref: MyNotificationTopic
```
and should be:
```
Value: !Ref MyNotificationTopic
```


rachfop:
Thank you for your feedback!

PR #612, opened 2020-03-04
Update aws-properties-ecs-taskdefinition-containerdefinitions.md

rpadovani:
The name of the container is actually a mandatory parameter.

If you do not specify it, creation of the Task resource will fail with this error:
> Container.name should not be null or empty. (Service: AmazonECS; Status Code: 400; Error Code: ClientException)

*Issue #, if available:*

*Description of changes:*


ckesser:
The "Type", "Required", "and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions.html#cfn-ecs-taskdefinition-containerdefinition-name) is updated after the specification is updated.

PR #617, opened 2020-03-09
Required should be Yes since it's a required field

ajdinhedzic-pfg:
The description says the DeploymentId is required by the required property is No.

*Issue #, if available:*

*Description of changes:*


ckesser:
Hi @ajdinhedzic-pfg , thanks for the feedback! The "Type", "Required", "and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). There's a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-stage.html#cfn-apigateway-stage-deploymentid) will be updated after the specification is updated. I'm closing this as a duplicate of #574.

PR #618, opened 2020-03-11
Update aws-resource-lambda-function.md

rainabba:
Update Tags property information and add Tags property examples that work with aws::serverless::function. Related issue: https://github.com/awslabs/serverless-application-model/issues/183

*Issue #, if available:*

*Description of changes:*


seittema:
@rainabba I don't think this edit is correct.  The Tag property for the AWS::Lambda::Function resource is correct in the current documentation as a resource tag.

The issue you are describing is with the AWS::Serverless::Function, which is a different resource.  I think the confusion lies in the explanation of the tags section located [here](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-tags) I believe that would be the correct spot to provide clarity

rainabba:
No doubt that my confusion landed me on the wrong docs. :)  I just did my search from scratch then used the "edit this page" option and did end up at https://github.com/awsdocs/aws-sam-developer-guide/blob/master/doc_source/sam-resource-function.md this time. I'll close this PR and start again over there.

rainabba:
After just a bit more thought, I realize something is still off because I and so many others have gone through this same confusion as evidenced at: [https://github.com/awslabs/serverless-application-model/issues/183](https://github.com/awslabs/serverless-application-model/issues/183)

I agree that is the right doc to reflect the information and just as I went to start the PR, I realized it IS the doc with the right information. In my past post on that issue thread, you can even find that I link to that doc as the one we should have landed on. Still, many of us did not and that means something else in the docs is off.

That prompted me to do an incognito search for "aws::serverless::function tags" to find [that](https://github.com/awslabs/serverless-application-model/issues/183) issue is the top search result.

This time, the 2nd hit was the German version of that page that IS causing confusion so I'm going to do a PR on that page ( [https://docs.aws.amazon.com/de_de/serverless-application-model/latest/developerguide/sam-resource-function.html](https://docs.aws.amazon.com/de_de/serverless-application-model/latest/developerguide/sam-resource-function.html) ) 

When I hit the "edit this page" button for the German version, it lands me at: [https://github.com/awsdocs/aws-sam-developer-guide/blob/master/doc_source/sam-resource-function.md](https://github.com/awsdocs/aws-sam-developer-guide/blob/master/doc_source/sam-resource-function.md)

Then I checked the US-English version ( https://docs.aws.amazon.com/en_us/serverless-application-model/latest/developerguide/sam-resource-function.html ) and hit its' "edit this page" button which ALSO brought me to the same github page. 

Looks like I need to figure out how AWS manages translations/languages in their repos now.

I just wanted to leave this finding for anyone else interested/affected.

seittema:
> 
> 
>In my past post on that issue thread, you can even find that I link to that doc as the one we should have landed on. Still, many of us did not and that means something else in the docs is off.
> 
> That prompted me to do an incognito search for "aws::serverless::function tags" to find [that](https://github.com/awslabs/serverless-application-model/issues/183) issue is the top search result.


I think part of the confusion is the overlap between CloudFormation and SAM. In my personal opinion, while SAM looks like CloudFormation, it actually isn't core CFT and in the end it's a transform that produces cloudformation.  That's why the SAM resources are not part of the CloudFormation documentation and instead their entirely own top level documentation "AWS Serverless Application Model".

I did a google search for the AWS::Serverless::Function resource and the first result was the proper [page](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html) 

Hopefully you will make the docs will be clearer on the SAM side to prevent the confusion for others

rainabba:
New PR at: https://github.com/awsdocs/aws-sam-developer-guide/pull/48

rainabba:
Unless you're searching in Incognito, that search isn't meaningful for others.

Here is an incognito search for `AWS::Serverless::Function resource`
image

When we realize our YAML for Tags isn't working, the next logical search is `AWS::Serverless::Function Tags` which nets:

image

Follow that 2nd link and what you end up finding is "AWS CloudFormation Compatibility: This property is similar to the [Tags](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-tags) property of an AWS::Lambda::Function. The Tags property in AWS SAM consists of Key:Value pairs. In AWS CloudFormation it consists of a list of Tag objects. When the stack is created, AWS SAM automatically adds a lambda:createdBy:SAM tag to this Lambda function and the corresponding IAM execution role" That link sends us to the resource page for "AWS::Lambda::Function".

Little doubt why I landed here, but I now see that my edits were quite the opposite of what they need to be to help this confusion from this side. Since someone decided to overly complicate life by using a different data-type for Tags in CloudFormation templates (say what you want about the translate, but it's all still CloudFormation in every sense and AWS does nothing to say it's "totally different" so this is a problem that's now become an issue for this resource.

That said, perhaps this doc should be updated to provide some hint (at Tags anyway) to help? I don't think I'm in any position to suggest that wording, but I hope someone else will given their proximity to this project and an appreciation for this confusion.

Personally, I think it's silly that the data-type difference was made, especially if there is a translation in the middle. :/

rainabba:
This still needs to be addressed and here's a complimentary issue (or a dupe edit via a different repo, it's been a while since I opened both): https://github.com/awsdocs/aws-sam-developer-guide/pull/48/

PR #631, opened 2020-03-20
Correct `Parameters` property, add info and example

mtboren:
The previous value for *Type* that was `JSON` is incorrect.  This can lead to quite a bit of wasted time for users trying to craft a proper value for the `Parameters` property. This file change adds a few clarifying statements about the value for `Parameters`, and adds a YAML example of the syntax for the sub-properties of the `Parameters` property

*Issue #, if available:*  none

*Description of changes:*  add detail and example for the `Parameters` property of this `MaintenanceWindowRunCommandParameters` property type, in order to help others quickly and easily add such a section to their CloudFormation template (the current documentation contents that said "JSON" as the value type was misleading, and there was no example)


rachfop:
Thank you for your suggestion.  The "Type", "Required", and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). Meaning, that the Parameters property cannot change from Json > Object (Object with properties that represent the parameters for the RUN_COMMAND task). I do appreciate the code example and will merge your suggestion. Thank you again for your feedback, it helps.

PR #642, opened 2020-04-02
Update LogDestination for StateMachine

seittema:
You can specify a LogDestination for any type of StateMachine (Express or Standard). Removing the Note that this only applies to Express


rachfop:
Hello @seittema, thank you for your suggestion. We have made changes per your suggestion. If you would like to rebase to remove conflicts please do so.

PR #650, opened 2020-04-15
Add note regarding StackSet operation concurrency

mpata:
*Description of changes:* Added a note on 4 pages of the User Guide to briefly explain how to set the values of MaxConcurrentCount and FailureToleranceCount to achieve deployment concurrency.
Hit this wall a long time ago and got a customer complaint about the same thing yesterday.

The [API Reference](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_StackSetOperationPreferences.html) explains it, but this information is buried and difficult to find.


rachfop:
Thank you for pointing this out. Adding that section from the API reference will help. Thanks.

PR #658, opened 2020-04-22
Update aws-resource-apigateway-basepathmapping.md

dongnorm:
https://docs.aws.amazon.com/cli/latest/reference/apigateway/create-base-path-mapping.html
--rest-api-id is marked as required.

*Issue #, if available:*

*Description of changes:*


rachfop:
The "Type", "Required", "and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) will be updated after the specification is updated.

PR #659, opened 2020-04-22
Update aws-resource-apigateway-basepathmapping.md

dongnorm:
Name of the API doesn't work, need to use ARN e.g. RestApiId: !Ref API

*Issue #, if available:*

*Description of changes:*


mcdwil:
Thanks for the PR! You're right that API's name doesn't work here. Just the ID, not the full ARN is needed. I've updated the description in the source.

PR #661, opened 2020-04-28
Add a page for the intrinsic `Condition` function

ps-jay:
*Issue #, if available:*
After discussion with @rjlohan 

*Description of changes:*
`Condition` is like `Ref` - but it's specifically for conditions.  Also like `Ref`, `Condition` does not require the `Fn::` prefix, and its' short form is `!Condition`.

Thus far `Condition` has been undocumented except for a small note and some code snippets in [`intrinsic-function-reference-conditions.md`](https://github.com/awsdocs/aws-cloudformation-user-guide/blob/master/doc_source/intrinsic-function-reference-conditions.md#condition-functions).

This PR is to give `Condition` its' own reference page.

----

### DIFF
--- a/doc_source/intrinsic-function-reference-condition.md
+++ b/doc_source/intrinsic-function-reference-condition.md
@@ -1,0 +1,63 @@
+# `Condition`<a name="intrinsic-function-reference-condition"></a>
+
+The intrinsic function `Condition` returns the evaluated result of the specified *condition*\.
+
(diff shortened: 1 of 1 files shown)


rachfop:
Hello @ps-jay thank you very much for putting this together. Your original contribution is very much appreciated as this feature of CloudFormation has been under documented for quite some time. Feel free to find, create, or even suggest more missing pieces of documentation in the future, as every piece of feedback helps.

PR #664, opened 2020-04-29
Update aws-resource-elasticsearch-domain.md

eul721:
*Issue #, if available:*

*Description of changes:*

Adding clarity to the use of `LogPublishingOptions` in `AWS::Elasticsearch::Domain`. Currently, it doesn't explain the full details on how to provide its proper values, other than "map of LogPublishingOptions"


aetter:
Hi @eul721, I'd rather add the new JSON to the first example so that we don't make the second example any longer than it already is, but I'll just merge and do some after-the-fact tweaks. Thanks a bunch for the contribution.

PR #667, opened 2020-05-01
s3 URL format

opstodevops:
amend website endpoints replacing . with - as documented https://docs.aws.amazon.com/general/latest/gr/s3.html#s3_website_region_endpoints

using endpoints in with . throws error "Tried to create an alias that targets <bucket-name>., type A in zone <zone-id>, but the alias target name does not lie within the target zone"

*Issue #, if available:*

*Description of changes:*


lincolahanbeck:
The S3 website endpoints have been updated to match those documented here: https://docs.aws.amazon.com/general/latest/gr/s3.html#s3_website_region_endpoints

PR #668, opened 2020-05-01
[BUGFIX] Removed invalid line break in Yaml code sample

afoeder:
This removed an incorrect use of a line break in the CloudFront
yaml example code snippet. Before the fix the code would result
in a compile(r) error.


rachfop:
Thank you for your submission! Every piece of feedback helps us direct our attention to where it is needed.

PR #671, opened 2020-05-04
Adding required tag for json structure

chetan-acharekar:
The fact that each value of mapping itself should be map, is missed. This will save valuable time of developers if they just set it to string, deploy then see error mentioning same.

*Issue #, if available:*

*Description of changes:*


rachfop:
Thank you for your submission! Every piece of feedback we receive helps us place our focus where it is needed the most.

chetan-acharekar:
:)

PR #672, opened 2020-05-04
Add a link for accepted Eventtypeids

abhinav-badola:
https://docs.aws.amazon.com/codestar-notifications/latest/userguide/concepts.html#concepts-api

*Issue #, if available:*

*Description of changes:*


rachfop:
Hello, thank you for pointing that out. The "Type", "Required", "and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) will be updated after the specification is updated.

PatMyron:
think this edit's just above those autogenerated fields

PR #673, opened 2020-05-04
Describe valid values for Amplify redirect/rewrite status

FergusMcGlynn:
The Amplify redirect/rewrite status value "404-200" isn't at all obvious and needs to be documented so that users who want to create a "404 rewrite" rule (common for React apps) can do so.

*Issue #, if available:* N/A

*Description of changes:* Added a table listing the valid values that `Status` can have when declaring a redirect/rewrite rule on an Amplify app. This is because the value for a 404 rewrite rule is not at all obvious, nor documented anywhere that I could find.

### DIFF
--- a/doc_source/aws-properties-amplify-app-customrule.md
+++ b/doc_source/aws-properties-amplify-app-customrule.md
@@ -42,6 +42,15 @@ To declare this entity in your AWS CloudFormation template, use the following sy
 
 `Status`  <a name="cfn-amplify-app-customrule-status"></a>
  The status code for a URL rewrite or redirect rule\.   
+ 
+|Value    |Represents                |
+|---------|--------------------------|
+|`200`    |200 (Rewrite)             |
+|`301`    |301 (Redirect - Permanent)|
+|`302`    |302 (Redirect - Temporary)|
+|`404`    |404 (Redirect)            |
+|`404-200`|404 (Rewrite)             |
+
 *Required*: No  
 *Type*: String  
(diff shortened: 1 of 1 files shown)


rachfop:
Hello, thank you for pointing that out. The "Type", "Required", "and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) will be updated after the specification is updated.

FergusMcGlynn:
Fair enough. My commit doesn't actually modify any of those fields, just the description for `Status`. As long as the documentation for the Amplify resource in Cloudformation gets updated to tell users that value "404-200" means "404 (Rewrite)" then I'm happy.

rachfop:
I'll reopen this pull request and have Kristian take a look.

FergusMcGlynn:
Thanks for looking into this. The pull request is still closed BTW - does it need to be reopened?

KristinaPeterson:
Thank you for this suggestion. The description for "status" comes from the Amplify API reference. I can work with the Amplify team to add the information you provided.

FergusMcGlynn:
That sounds good!

KristinaPeterson:
I apologize for the delay in getting this information published. It is now live in both the Amplify CloudFormation and Amplify API documentation. Thank you again for the suggestion.

PR #678, opened 2020-05-11
Create-Stack on CLI does not upload to S3

taylorb-syd:
*Issue #, if available:* N/A

*Description of changes:*

The documentation currently claims that templates uploaded via the CLI will be duplicated into an S3 bucket in the customers account. This is not true, and _only_ applies to uploads using the Web Console. Therefore having this note on this page is misleading.

Removing the note.


rachfop:

PR #680, opened 2020-05-12
Add information about behaviour of DB Cluster state on DB Cluster Par…

margod:
Add information about behaviour of DB Cluster state on DB Cluster Parameter Group updates

*Issue #, if available:*

*Description of changes:*
Clarification for those who scratch head why DB Cluster Parameter Group update seems stuck


randyurbano:
Thanks for your feedback. I confirmed that your update is correct.

PR #683, opened 2020-05-18
Spelling Correction on InstanceType Anchor

jsravencroft:
*Issue #, if available:*

*Description of changes:*
It's just a correction of spelling errors in an anchor.  Maybe it's no big deal, sorry for the trouble.


rachfop:
@jsravencroft thank you for your submission. That anchor link is a valid, so we might have to change a few files. Good catch!

rachfop:
The "Type", "Required", and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value (update the url) in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) will be updated after the specification is updated.

PR #693, opened 2020-05-22
Update aws-resource-directoryservice-microsoftad.md

rachfop:
The [API Create Alias](https://docs.aws.amazon.com/directoryservice/latest/devguide/API_CreateAlias.html) indicates that it is calling for a string and not a Boolean.

*Issue #, if available:*

*Description of changes:*


rachfop:
The "Type", "Required", "and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) will be updated after the specification is updated.

PR #700, opened 2020-05-27
Fix Fn::Join docs example

mluttrell:
*Issue #, if available:*

*Description of changes:*

The Fn::Join docs include an example with a reference to "Partition".  The correct pseduo parameter is "AWS::Partition".  Adding this example to a CloudFormation template will cause the template to fail validation.

Pseudo parameter reference here:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html


rachfop:
Hello, thank you for your suggestion.

PR #705, opened 2020-05-31
Removed TCP and UDP warning on same port

dsalamancaMS:
I have removed the note that states that an error will be thrown if a task definition is created on both TCP and UDP ports, as this is no longer true

*Issue #, if available:*

*Description of changes:* I have removed the note that states that an error will be thrown if a task definition is created on both TCP and UDP ports, as this is no longer true


joelbrandenburg:
Thanks for your contribution!

PR #706, opened 2020-06-02
Update aws-properties-elasticsearch-domain-cognitooptions.md

dongnorm:
According to the Prerequisites of Amazon Cognito Authentication for Kibana - Prerequisites - https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-cognito-auth.html#es-cognito-auth-prereq

If Enabled, you need all 3 of those properties set, otherwise in CFN you get an "xxx must be specified" error. By saying "Required: No", the document does not clearly indicate the necessity of those properties, which is misleading the users into that error.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/doc_source/aws-properties-elasticsearch-domain-cognitooptions.md
+++ b/doc_source/aws-properties-elasticsearch-domain-cognitooptions.md
@@ -36,18 +36,18 @@ Whether to enable or disable Amazon Cognito authentication for Kibana\. See [Ama
 
 `IdentityPoolId`  <a name="cfn-elasticsearch-domain-cognitooptions-identitypoolid"></a>
 The Amazon Cognito identity pool ID that you want Amazon ES to use for Kibana authentication\.  
-*Required*: No  
+*Required*: Yes if enabled  
 *Type*: String  
 *Update requires*: [No interruption](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-no-interrupt)
 
 `RoleArn`  <a name="cfn-elasticsearch-domain-cognitooptions-rolearn"></a>
 The `AmazonESCognitoAccess` role that allows Amazon ES to configure your user pool and identity pool\.  
-*Required*: No  
+*Required*: Yes if enabled  
 *Type*: String  
 *Update requires*: [No interruption](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-no-interrupt)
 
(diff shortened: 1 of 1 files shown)


aetter:
Hi @dongnorm, these aren't required in the same way that EBSOptions aren't required, DedicatedMasterCount isn't required (only if DedicatedMasterEnabled is true), etc.

dongnorm:
Hi @aetter, if you say so - then could you please provide a working example template, where CognitoOptions Enabled is true, and either one of IdentityPoolId, RoleArn and UserPoolId is not specified?

Per my testing, it is not going to work, you NEED all 3 of them, which makes them required. Happy to be corrected.

aetter:
Correct, but that's not the point. You can make valid templates that don't include those options. Ergo, optional.

dongnorm:
Do you think it is OK if a user comes to this document page, sees "Required: no" and makes a template without one of those properties then proceeds to see an error?

Do you think it is a great user experience?

Do you think it is not confusing at all, for an average user, who might be cross referencing the other document I linked, and wonder "why is it not required in CFN when Cognito document says these are required?"

Do you think it is more common for a user who lands on this page, to end up writing:
blah
CognitoOptions:
  Enabled: false
blah

Which is unnecessary, because they can just skip it entirely.

Or would they actually want to use it so they will actually set Enabled to true? And now they think "oh I don't need to specify any of the rest because they are not required."

ajmay-aws:
So, this sounds like a case where the properties should be marked as Required: Conditional. This is how we mark properties in the documentation where the property is only required under certain conditions (usually if another property is set to a certain value.)

So, we'd mark all three properties as Required: Conditional, and then add the following to each property description to make it clear to the user what the condition is:

"If you set Enabled to true, you must specify values for IdentityPoolId, RoleArn, and UserPoolId."

Also, I'm assuming that, since Enables is not required, the default if it is not specified is false? If that's correct, we should probably add that to the description of the Enabled property as well.

@dongnorm Would that address your issue?

ajmay-aws:
And to Andrew's point, the properties need to remain defined as required = false in the resource spec, since from the perspective of validating a stack template, those properties need not be present in the resource definition.

dongnorm:
@ajmay-aws thanks, I knew there was something we use to describe this kind of attributes but I couldn't recall. Conditional suits this scenario perfectly.

I have updated the PR, let me know if further edits are required?

swiercek:
Content is service writer owned. Leaving open for review.

joshbean:
Closing this issue or pull request in advance of archiving this repo. For more information about the decision to archive this repo (and others in the "awsdocs" org), see [the announcement on the AWS News Blog](https://aws.amazon.com/blogs/aws/retiring-the-aws-documentation-on-github/).

PR #710, opened 2020-06-04
Delete false ordering requirement

edenhochbaum:
This statement seems unlikely to actually be true, just based on the json spec ("An object is an unordered set of name/value pairs")

*Issue #, if available:*

*Description of changes:*

Removed false statement


rachfop:
Thank you for your suggestion. You are correct and I will be merging your pull request.

PR #714, opened 2020-06-10
Renamed misleading title

WingsLikeEagles:
Title was misleading.  This page is about using an Existing Template which describes a resource.  It's not about using an Existing Resource.  It has no information on how to create that template from a resource.


ckesser:
Thank you for your feedback, @WingsLikeEagles . This topic describes how to make a stack out of existing resources, using a template that defines those resources. The sample template provided in this topic serves as a guide for defining the existing resources that will be imported into the new stack. I'm going to close this request; the title as-is more accurately describes the procedure in this topic. If you have feedback for the body of this topic, please let us know.

PR #717, opened 2020-06-11
Update aws-resource-cloudwatch-compositealarm.md

amzcarlli:
AlarmActions in template example (json and yaml) should be "list of strings"

*Description of changes:*

The example template is inconsistent with Properties description.

"AlarmActions" in template example (json and yaml) should be "list of strings" instead of "strings", and using the template example will results in "Internal Failure" when creating AWS::CloudWatch::CompositeAlarm due to incorrect data type.


rachfop:
Hello, thank you for your feedback.

PR #720, opened 2020-06-18
Updated count (was 2, updated to 3)

volkanpaksoy:
I think this bit was missed when CODE_DEPLOY was added

*Issue #, if available:*

*Description of changes:*


rachfop:

PR #722, opened 2020-06-18
Added a note about valid values for SSL Mode when using S3 as the engine type

DavidTPate:
*Issue #, if available:* N/A

*Description of changes:* When setting up CloudFormation for DMS endpoints with S3 I ran into an issue where I wasn't able to use the values `require`, `verify-ca`, or `verify-full` when using the S3 engine type. I'm not able to find this mentioned in any other documentation around DMS so I'm either missing the paragraph about it or it needs to be added to other docs too. I found this out by experimenting with values.


rachfop:
Thank you for your suggestion. Every piece of feedback helps our documentation.

PR #723, opened 2020-06-18
Updated required statement for external table definition

DavidTPate:
*Issue #, if available:* N/A

*Description of changes:* While doing setup with DMS today I found through using CloudFormation and the APIs for DMS that the External Table Definition parameter is required. I haven't been able to find another piece of documentation that states this but it likely needs to be updated elsewhere as well.


rachfop:
The "Type", "Required", and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) will be updated after the specification is updated.

rachfop:
Thank you for your submission.

PR #725, opened 2020-06-20
Update argument name, from --stackname to --stack-name

fedecarg:
Update argument name


rachfop:

PR #731, opened 2020-06-26
Adds info logical name being case-sensitive

### LINKED ISSUE #729: Resource Logical ID is case sensitive, but undocumented
In templates, the resource logical ID is case-sensitive, but documentation does not make this clear.

ttres:
This change includes the information that a resource logical name is case-sensitive, so 'IAMrole' and 'IamRole' are different resources, per example, and the stack will fail with 'Template format error: Unresolved resource dependencies' when trying to use the names interchangeably.

Fix #729 

### DIFF
--- a/doc_source/intrinsic-function-reference-ref.md
+++ b/doc_source/intrinsic-function-reference-ref.md
@@ -2,7 +2,7 @@
 
 The intrinsic function `Ref` returns the value of the specified *parameter* or *resource*\.
 + When you specify a parameter's logical name, it returns the value of the parameter\.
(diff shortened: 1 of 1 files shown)


rachfop:
Hello @ttres thank you for taking interest in the documents! Every piece of feedback helps. By definition, [Logical Names](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/glossary.html#logical-name) are case sensitive. 
````
logical name:

    A case-sensitive unique string within an AWS CloudFormation template that identifies a resource, mapping, parameter, or output. In an AWS CloudFormation template, each parameter, resource, property, mapping, and output must be declared with a unique logical name. You use the logical name when dereferencing these items using the Ref function.
````
I can take a look into seeing if that can be cleared up in the documents. Thanks again for the suggestion.

PR #732, opened 2020-06-26
Mark "memory" and "vcpus" as required

ozydingo:
*Issue #, if available:*

For AWS::Batch::JobDefinition ContainerProperties, the parameters `Memory` and `Vcpus` are marked as `Required: No`. However, stack deploys fail if these parameters are not set.

*Description of changes:*

Mark these parameters as required.


HLeeHart:
Memory and Vcpus must be specified for every node, but multi-node parallel jobs can specify them in multiple places. They are only required to be specified in one. If you look at the example multi-node parallel job definition, you'll see that the containerProperties (parent to memory and vcpus) is not even specified in the top of the job definition, but the container member of the nodeRangeProperties member does contain those members.
https://docs.aws.amazon.com/batch/latest/userguide/example-job-definitions.html#example-mnp-job-definition
I added this sentence to the description for memory and vcpus: "This is required but can be specified in several places for multi-node parallel (MNP) jobs; it must be specified for each node at least once."
You can see that sentence in the docs at https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-containerproperties.html#cfn-batch-jobdefinition-containerproperties-memory 
GitHub will be updated in the next few days.

PR #734, opened 2020-07-01
Update aws-properties-ecs-service-deploymentcontroller allowed values

snicol-indeed:
The current page does not include `CODE_DEPLOY` as an allowed value. This is inconsistent amongst other pages such as https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DeploymentController.html. I have edited the page to include `CODE_DEPLOY` as an allowed value.


rachfop:
Hello, @snicol-indeed thank you for the suggestion. The docs are updated.

PR #738, opened 2020-07-10
Replaced the obsolete db.m1.small class

iamgeromep:
*Issue #, if available:*

*Description of changes:*
Replace “db.m1.small” to “db.t2.small”
m.1small instance is already obsolete and a previous generation type

Reference:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/quickref-rds.html


rachfop:
Hello, thank you for your suggestion. I can confirm that the m1 class has been changed to t2 per https://aws.amazon.com/rds/previous-generation/

PR #740, opened 2020-07-13
Replaced the obsolete db.m1.small class

carlo-acebedo:
Issue #, if available:

Replaced "db.m1.small" to "db.t2.small"
RDS no longer support the creation of db.m1.small instance type.

Reference:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-dependson.html


rachfop:
Hello, thank you for your suggestion. I updated the YAML as well.

PR #742, opened 2020-07-15
AWS::EC2::SecurityGroup Ingress: Fix documentation

jmkgreen:
SourceSecurityGroupOwnerId - no longer has these restrictions

*Issue #, if available:*

*Description of changes:*
Tried following the docs and my changes were rejected until I added these fields.


julieso:
Thanks for the contribution. The peering VPC example on this page sets the port and protocol in addition to this property. I am not sure whether setting the IP address range is allowed. I will investigate and update the doc as needed.

PR #743, opened 2020-07-15
Fix broken Ref link in RotationLambdaARN section

drkmullins:
*Description of changes:*
Found a little `git status` surprise in the middle of this Ref link :) I enjoyed it, but my browser did not. Corrected it to https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ref.html


rachfop:
Thank you for finding that. We appreciate every contribution to our docs.

PR #748, opened 2020-07-21
Fix contradicting information

Feez:
Or perhaps it should be "Required: *maybe*"

*Description of changes:*
The description states "This is a required field". Immediately after, we see "Required: No". Fixed that


chrisdr:
You are correct and I approve and merge this request; however, that "Required: No: is autogenerated by a quirk in our system, and this doc will likely revert to the current state when the whole guide is again updated.  This quirk may be fixed sometime in the future.

PR #749, opened 2020-07-27
Clustered vs Standalone Outputs

austinpilz:
In the 'Return Values' section, there was no mention of which return values could be expected with what cluster mode. Cluster Mode Enabled, for example, only had the configuration endpoint, as where the Cluster Mode Disabled includes the primary and reader endpoints. The naming of the outputs doesn't make it immediately clear to a new user which outputs to expect, so these headers I think should give some quick direction.

*Issue #, if available:*

*Description of changes:* Modified the return values section to have headers to indicate which outputs are returned with each cluster mode of the elasticache replication group.

### DIFF
--- a/doc_source/aws-resource-elasticache-replicationgroup.md
+++ b/doc_source/aws-resource-elasticache-replicationgroup.md
@@ -384,12 +384,16 @@ For more information about using the `Fn::GetAtt` intrinsic function, see [Fn::G
 
 #### <a name="aws-resource-elasticache-replicationgroup-return-values-fn--getatt-fn--getatt"></a>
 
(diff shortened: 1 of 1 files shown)


austinpilz:
@rachfop Any update on this?

rachfop:
@austinpilz Thank you for your suggestion. This is an improvement to the documentation for sure. Due to the automation of our docs -- this change is a little more complex. I was unable to make your recommendation happen but I am still looking for a solution.

rachfop:
@austinpilz II'll add it to the team's list of items for future consideration and then close this pull request.

PR #752, opened 2020-07-28
Update aws-properties-cloudtrail-trail-eventselector.md

dongnorm:
If IncludeManagementEvents is false, DataResources must not be empty, otherwise we see an error: Invalid combination of parameters in EventSelectors. When includeManagementEvents is false, DataResources must not be empty.

Effectively, if IncludeManagementEvents is false, then DataResources is required and cannot be empty either.

*Issue #, if available:*

*Description of changes:*


rachfop:
The "Type", "Required", and "Update behavior" fields for each property are auto-generated from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). I've submitted a ticket to correct this value in the resource spec. The [source documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) will be updated after the specification is updated.

rachfop:
Scratch that, in this case, conditional can be added. I have assigned @GitGaby to this pull request to confirm. Thank you for your contribution!

rachfop:
Thank you for your suggestion. Every piece of feedback we receive helps us improve our documentation.

PR #756, opened 2020-07-30
Fix syntax errors in DBCluster YAML examples

kylelaker:
GitHub complained that there are inconsistent line endings in the file.


rachfop:
Thank you for your submission. Every piece of feedback we receive helps us put our focus where it is needed the most.

Issue #512, opened 2019-11-04
AWS CloudFormation limit update


taraslayshchuk:
To have all things in a same place. Could you please add information about [stack_name length limit](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-using-console-create-stack-parameters.html#cfn-using-console-create-stack-parameters-awstypes) and [tags name|value length limit](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html#w2ab1c21c10d212c13c17) into [AWS CloudFormation limits](https://github.com/awsdocs/aws-cloudformation-user-guide/blob/master/doc_source/cloudformation-limits.md)?

PatMyron:
[Output description length limit](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/outputs-section-structure.html)
and 1000 Change Sets per CloudFormation stack version as well

taraslayshchuk:
Nice! Well done! Great job! Thanks! @rachfop @PatMyron What about [tags name|value length limit](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html#w2ab1c21c10d212c13c17)?

PR #758, opened 2020-07-31
Add stack name size limit

frodera:
*Issue #, if available:* #512 

*Description of changes:* Maximum length for other resources' names (outputs, parameters, etc) are already
present in the document, the stack name should too for easy reference IMO.


rachfop:
Thank you for your contribution.

PR #760, opened 2020-08-03
Fix YAML

ritesh:
I think the YAML example is incorrect

*Issue #, if available:*

*Description of changes:*


rachfop:
Hello @ritesh Good catch! You're right, that YAML example shouldn't be formatted like that. Thank you for your contribution.

PR #766, opened 2020-08-07
Correct syntax error

pizza0rodeo:
Per the [BackupPolicy documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-efs-filesystem-backuppolicy.html), BackupPolicy contains a single value, not a list of values, so there should be no hyphen. If you use it with a hyphen, you get a "Model validation failed" error from CloudFormation. Without the hyphen, it works as expected.

*Description of changes:*
Removed the hyphen at the start of BackupPolicy's "status" property.


rachfop:
Thank you for correcting that and helping ensure the accuracy of our documentation.

PR #771, opened 2020-08-19
Update TLS section of cloudfront distribution docs

robertd:
Reference: https://github.com/aws/aws-cdk/pull/9738


joshbean:
Hi @robertd – Thanks for drawing our attention to this. As the CloudFront console says, we absolutely recommend the latest security policy (TLSv1.2_2019). This CloudFormation documentation got out of date when the 2019 security policy was added. To avoid this problem of the docs going out of date, I recently updated the CloudFront API reference documentation to remove the recommendation altogether (see [MinimumProtocolVersion](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_ViewerCertificate.html#cloudfront-Type-ViewerCertificate-MinimumProtocolVersion) in the CloudFront API docs). I'm inclined to do the same thing here with the CloudFormation documentation. That way, we reduce the risk of the docs potentially getting out of date again, if we ever release another security policy. As a result, I'm going to close this PR, and update the CloudFormation docs to match the CloudFront API reference docs.

Thanks again for bringing this to our attention, though! Certainly it was out of date to recommend the 2018 security policy, and something we needed to fix. I appreciate you drawing my attention to it 😃

robertd:
@joshbean You're welcome 😄

PR #772, opened 2020-08-25
docs: Change the RefreshTokenValidity precision

nhammond101:
According to https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-with-identity-providers.html, the RefreshTokenValidity is set in seconds, with a minimum of 600 seconds

*Issue #, if available:*

*Description of changes:*
Minor change to the timeframe precision of the RefreshTokenValidity from days to seconds, and the minimum set to 600, as described in other documentation


nhammond101:
Withdrawing PR as the Cloudformation docs are correct already

PR #773, opened 2020-08-26
Remove Fn::Transform short form

kkl-acies:
It isn't working

---

The `FN::Transform` isn't working/supported(?) so let's remove it. Ex: https://forums.aws.amazon.com/thread.jspa?threadID=321064


rachfop:
Thank you for bringing this up.

PR #777, opened 2020-08-31
AWS::CloudWatch::Alarm minumum period not suported

lcscarvalho:
when using the value `1` the following error is returned: 

```
Period must be 10, 30 or a multiple of 60 (Service: AmazonCloudWatch; Status Code: 400; Error Code: ValidationError; Request ID: [Request ID]) 
```

### DIFF
--- a/doc_source/aws-properties-cw-alarm.md
+++ b/doc_source/aws-properties-cw-alarm.md
@@ -198,7 +198,7 @@ The period, in seconds, over which the statistic is applied\. This is required f
 For an alarm based on a math expression, you can't specify `Period`, and instead you use the `Metrics` parameter\.  
 *Required*: No  
 *Type*: Integer  
-*Minimum*: `1`  
+*Minimum*: `10`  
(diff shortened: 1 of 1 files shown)


chrisdr:
HI Patrick,

This pull issue is for the Period value on this page<https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cw-alarm.html#aws-properties-cw-alarm-properties>, where my text accurately lists the actual valid values, while the auto-generated minimum shows 1 ,  which comes from the resource spec.  Or in this case maybe straight from the API service model, as the API reference<https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricAlarm.html#API_PutMetricAlarm_RequestParameters> has the same issue.

It sounds from a previous incident like this that we are being more forceful with filing tickets on the service team to have them fix the API/resource spec, right?   In this case, where the original error is in the API service model, does the service team need to fix only that, or do they also need to fix the CFN resource separately?

And how do I address the pull issue in Github in this case?   Merge it for now, but mention that the fix will likely be overwritten when the whole guide is updated the next time, and that the real fix must be done by the product team and that I have filed a ticket for them to do so?  Do we have a pre-approved wording for this situation?

Thanks - Chris

chrisdr:
Good catch, thanks

PR #781, opened 2020-09-09
Update aws-properties-apitgateway-stage-methodsetting.md

takanabe:
According to https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-request-throttling.html, `ThrottlingBurstLimit` is defined as follows:

>In API Gateway, the burst limit corresponds to the maximum number of concurrent request submissions that API Gateway can fulfill at any moment without returning 429 Too Many Requests error responses.

the burst limit does not define the maximum number of requests **per second** but **at any moment** .  This means API Gateway returns 429 when it receives more requests during a millisecond time-window than the value of `ThrottlingBurstLimit`.  So, the current explanation of `ThrottlingBurstLimit` is not accurate and a replacement with accurate documents make sense.


takanabe:
Let me close this issue since I don't think this PR will be merged... Please open this PR if necessary.

PR #783, opened 2020-09-09
Update resource-import-nested-stacks.md

udevasia:
*Issue #, if available:*
Step 6.a mentions that "an AWS::CloudFormation::Stack resource can be identified using the StackName property.". However, the Identifier property is StackId. The Identifier value has to be stack arn, instead fo stack name as specified in 6.b

*Description of changes:*
updated the information


rachfop:
Thank you for your suggestion. I can confirm that the Stack resource is identified by the property StackId not stackname. I will look into updating the docs per your suggestion.

PR #787, opened 2020-09-15
Remove note regarding multiple VPCs per HZ

JakeHendy:
*Description of changes:*

The examples and syntax state that multiple VPCs are supported when creating a private hosted zone, so the warning negating this should be removed.


joshbean:
Thanks for your PR. You're exactly right, this note shouldn't be there. I'll merge the PR and will get the documentation website updated soon.

PR #790, opened 2020-09-21
Fix doc for ApiGateway::Method "AuthorizationType" and "Integration" property

bigghe:
Fix "Required" attribute for ApiGateway::Method AuthorizationType property.
I think that in this way it's correct and more clear.


bigghe:
Sorry but I don't know why the diff in files change it's so messy.

PR #798, opened 2020-09-28
Fix indentation in "AWS::ApiGatewayV2::Route"

j0k3r:
*Description of changes:*
Fix indentation in example.
`cfn-lint` generated that error with the previous example:

> E1022 Join should be an array of 2 for Resources/Route/Properties/Target/Fn::Join


mcdwil:
Thanks!

PR #800, opened 2020-09-30
Fix key, value in an example of --key option

kit494way:
*Issue #, if available:*

*Description of changes:*
In example of `--key` option, key and value is reversed.


rachfop:
Thank you for your feedback and your pull request looks good. I think you're right that it should be reversed.

PR #807, opened 2020-10-02
Update tag value constraints

fredericbarthelet:
*Description of changes:*

Aligning share resource `Tags` property value constraints to schema described in https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_Tag.html

The existing `Value` description and constraints seems like a copy/paste error from `Key` description.

Sources of description used for this correction :
- CloudFormation API Reference documentation : https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_Tag.html
- Resource Groups & Tagging API Reference documentation : https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_Tag.html

Actual behavior has been tested in https://github.com/serverless/serverless/issues/8304  

### DIFF
--- a/doc_source/aws-properties-resource-tags.md
+++ b/doc_source/aws-properties-resource-tags.md
@@ -40,7 +40,7 @@ The key name of the tag\. You can specify a value that is 1 to 128 Unicode chara
 *Type*: String
 
 `Value`  <a name="cfn-resource-tags-value"></a>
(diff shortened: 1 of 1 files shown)


rachfop:
Thank you for your pull request. It does appear that it was a copy/paste error. I am glad you could find and share an updated description for the tag property.

PR #812, opened 2020-10-07
Update cfn-lambda-function-code-cfnresponsemodule.md

zanderz:
The response module source code for Python 2 functions was missing a very important call to urllib2.urlopen(), and declaration of response, without which it does not work.
There was also a bug in the line that prints the result.

*Issue #, if available:*

*Description of changes:*


rachfop:
Thank you for your suggestion. I am glad you were able to get the py2 code working. I'll merge your changes.

PR #814, opened 2020-10-09
Update aws-resource-config-conformancepack.md

seemuellera:
Clarification that the Name of the bucket needs to specified and not the ARN

*Issue #, if available:*

*Description of changes:*


JimbroAmazon:
Assigning to Shubhangi Vajpayee (writer for AWS Config)

seemuellera:
Good morning,

I just want to follow up on this PR. It is probably outdated by now but I would like to know if it makes sense in general. I think the documentation of the parameters should be clear in the values they accept. Not knowing if you have to specificy the bucket's name, ARN, or S3 URL leads to unnecessary try-out loops and frustration.

PR #817, opened 2020-10-12
Fix lambda permission function reference

Graham-M:
*Issue #, if available:*
N/a

*Description of changes:*
I believe the reference to the lambda function is incorrect in the YAML context.


rachfop:
Thank you for your suggestion. Every piece of feedback helps.

PR #820, opened 2020-10-16
roles names *must* match

felipe1982:
MUST is stronger than SHOULD. Should implies a suggestions which may be ignored.

*Issue #, if available:*

*Description of changes:*


rachfop:
Thank you for your feedback.

Issue #797, opened 2020-09-24
AWS::Cognito::UserPool.CaseSensitive is required


xrn:
Template
```
{
  "AWSTemplateFormatVersion": "2010-09-09",
  "Resources": {
    "myCognito": {
      "Properties": {
        "UsernameConfiguration": {}
      },
      "Type": "AWS::Cognito::UserPool"
    }
  }
}
```

Error
```
Value null at 'usernameConfiguration.caseSensitive' failed to satisfy constraint
```

Current documentation https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cognito-userpool-usernameconfiguration.html

Looks like CaseSensitive is required

xrn:
@prazmaap what do you think here?

douglas-gibbons-lark:
@prazmaap do you want me to create a PR for you?

rachfop:
Our tools auto-generate the "_Required_", "_Type_", "and "_Update requires_" fields for each property from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). 
I submitted a ticket to correct this value. We will refresh the [documentation ](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html)after the specification update.

xrn:
@rachfop seems that this is still not requried - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cognito-userpool-usernameconfiguration.html

PR #824, opened 2020-10-19
Change AWS::Cognito::UserPool.CaseSensitive Required to Yes

xrn:
*Issue #, if available: #797 

*Description of changes:*


xrn:
@rachfop any chance for merge?

AMZ-brandon:
Hi @xrn, thanks for the pull request! And apologies for the late response. That field is set at the service level rather than the docs level. However, my understanding is that because `CaseSensitive` has a default value of _True_, you don't need to specify it, or else it wouldn't be a default.
If your experience doesn't track with that, can you provide more information about that? Thanks!

PR #827, opened 2020-10-21
Update what-is-cfnstacksets.md

Santiago-Freitas:
Add information in which regions StackSets is supported. Currently we don't have anywhere in the StackSets docs a list where it's available. This makes hard for customers and field to find where the feature is supported.

*Issue #, if available:*

*Description of changes:*
Add information in which regions StackSets is supported. Currently we don't have anywhere in the StackSets docs a list where it's available. This makes hard for customers and field to find where the feature is supported.


rachfop:
This is information that should definitely be included in the StackSets doc. Thank you for pointing it out.

PR #839, opened 2020-10-31
Update kinesisfirehose deliverystream format conversion

Nxtra:
*Description of changes:*

Update of the example for firehose record format conversion. Includes:
- formatting
- updated properties of resources
- update json in yaml formatting

I validated this yaml through a successful deployment in a cf stack.

### DIFF
--- a/doc_source/aws-resource-kinesisfirehose-deliverystream.md
+++ b/doc_source/aws-resource-kinesisfirehose-deliverystream.md
@@ -261,178 +261,158 @@ The following example shows record format conversion\.
 
 ```
 AWSTemplateFormatVersion: 2010-09-09
-            Description: Stack for Firehose DeliveryStream S3 Destination.
-            Resources:
-            
-            GlueDatabase:
-            Type: AWS::Glue::Database
-            Properties: 
-            CatalogId: !Ref AWS::AccountId
-            DatabaseInput: {}
-            
-            GlueTable:
-            Type: AWS::Glue::Table
-            Properties:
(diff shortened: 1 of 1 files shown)


Gaianab:
thanks for your feedback, I made the requested changes.

Nxtra:
@Gaianab , thanks for the feedback.
Why didn't you just merge my PR? 
- I now have the feeling my work disappeared. 
- You would have had it properly formatted in the PR:

image

Gaianab:
Hi! So first of all, thank you so much for your feedback! The way this works is when we open source our AWS docs, we generate markdown files based on the actual source files that are behind our aws docs. when someone does a pull request here, we review the feedback and if we agree with it (99.9% of the time) we made the change directly in the source files, then we generate new markdown files and upload them here. and then we either merge the PR or just close it. either action is purely nominal. what i've done is fixed the formatting in the source files and republished it - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kinesisfirehose-deliverystream.html

then in a few days, a team that owns Cloud Formation docs and open source CLoud Formation docs will generate new markdowns based on my (yours) changes and push them here.

you work hasn't disappeared. it'll show up on github in a few days. and the proper formatting is already live on https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kinesisfirehose-deliverystream.html 

I hope this makes sense, it's just the process we have to follow for open sourcing. 

Thanks again!

PR #851, opened 2020-11-13
Update aws-resource-fsx-filesystem.md

shantgup:
StorageCapacity is required, will throw error like below without it.

```
1 validation error detected: Value null at 'storageCapacity' failed to satisfy constraint: Member must not be null (Service: AmazonFSx; Status Code: 400; Error Code: BadRequest; Request ID: 852ac170-3486-4cba-9b91-8e029d7359fc; Proxy: null)
```

*Issue #, if available:*

*Description of changes:*


rmarcuce (approved):
The "Type", "Required", "and "Update behavior" fields for each property are auto-generated from the resource specification. I've submitted a ticket to correct this value in the resource spec. The source documentation will be updated after the specification is updated.

swiercek:
Content was updated. Closing PR.

PR #852, opened 2020-11-13
Update aws-properties-fsx-filesystem-windowsconfiguration.md

shantgup:
Either `ActiveDirectoryId` or `SelfManagedActiveDirectoryConfiguration` is required, so they should be conditional.

*Issue #, if available:*

*Description of changes:*


rmarcuce:
The "Type", "Required", "and "Update behavior" fields for each property are auto-generated from the resource specification. I've submitted a ticket to correct this value in the resource spec. The source documentation will be updated after the specification is updated.

PR #860, opened 2020-11-17
Add missing requirement for SNS FIFO topic name

trina24:
*Description of changes:*

Added important information about required `.fifo` suffix for Simple Notification Service FIFO topic names.
An attempt to create FIFO topic according to the rules currently described in user guide (suffix is not mentioned) causes `InvalidParameterException` with undescriptive message `Invalid parameter: Topic Name`.


carlasp (approved):
Thanks for letting us know about this!

PR #872, opened 2020-12-03
Added detail that AllowedPattern parameter argument is required to ma…

magnetikonline:
*Issue #, if available:*

*Description of changes:*

- Property `AllowedPattern` which defines a regular expression match for string values must always match the _entire_ value provided. I think it's worthwhile to note this on the documents as it's tripped me up before.
- Chomped out some trailing whitespace on save, I'm _assuming_ this whitespace is not needed for document generation? If it is required, can re-create commit.

Thanks.


rachfop:
Thank you for your feedback on our documentation. I will commit your pull request, but some of the white space is auto-generated and may be overwritten in a future commit. Thank you for providing an improved description for for `AllowedPattern.`

PR #874, opened 2020-12-04
Update the required property from No to Yes for the Metric name of the CloudWatch Alarm Metric

ChrisPearce:
Metric name is a required field but the Required property incorrectly displayed as No rather than Yes

*Issue #, if available:*
None raised

*Description of changes:*
Change the required property of the MetricName to Yes rather than No


ChrisPearce:
Hi. Is there any feedback on this proposed pull request?

rachfop:
Our tools auto-generate the "_Required_", "_Type_", "and "_Update requires_" fields for each property from the [resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html). 
@chrisdr has submitted a ticket to correct/investigate this value. We will refresh the [documentation ](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html)after the specification update.

PR #875, opened 2020-12-04
ArchiveName is not a valid property

Nr18:
*Issue #, if available:*

*Description of changes:*

While copying the sample I notices that the sample included an `ArchiveName`, but the properties section does not mention the `ArchiveName` property and `cfn-lint` also complains about it.

### DIFF
--- a/doc_source/aws-resource-events-archive.md
+++ b/doc_source/aws-resource-events-archive.md
@@ -90,7 +90,6 @@ The following example creates an archive for all EC2 events sent from the defaul
   "SampleArchive": 
     "Type" : "AWS::Events::Archive",
     "Properties" : {
-        "ArchiveName" : "MyArchive",
         "Description" : "Archive for all EC2 events",
         "EventPattern" : {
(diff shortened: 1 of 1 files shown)


randalld-aws:
Thanks for pointing this out. ArchiveName is actually valid and you can use it to create a resource. But we seem to have a discrepancy in the public resource spec and public docs. Under investigation.

Nr18:
👌I only removed it because it was not in the spec. 

So if it's coming depending on when? You might want to remove the confusion from the public docs and add it when it is available.

rachfop:
@Nr18 ArchiveName is back in the [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-archive.html) and resource spec: 
```
 "Attributes": {
        "ArchiveName": {
          "PrimitiveType": "String"
        },
        "Arn": {
          "PrimitiveType": "String"
        }
      }
```

Thank you for bringing this to our attention.

PR #878, opened 2020-12-07
Add ArchiveName

Adam21e:
Add missing ArchiveName to the Properties section.

*Issue #, if available:*

*Description of changes:*

 Just adds the ArchiveName property.


randalld-aws:
Thanks for reporting this. It is an issue with the public resource spec rather than docs. I can't fix this unless the public resource spec is updated to also include it. Update pending.

PR #882, opened 2020-12-09
Correct schema used in uri field

kchandra17:
cfn-init performs a comparison against schemaless URIs, so these examples do not work as written.

*Issue #, if available:*

*Description of changes:*


rachfop:
Thank you for your feedback, this update is pending investigation. Updates will be given in this thread.

rachfop:
Thank you for your suggestion and improving the docs.

PR #893, opened 2020-12-26
fixed typo of items field for cloudfront keygroup

enable887:
Changed the configuration field 'items' to 'Items', using the lowercase one will cause Cloudformation to fail.


joshbean:
Thanks for the PR! It looks like we have an issue in one of our spec files where the "i" is incorrectly lowercase, which is generating this documentation with a lowercase "i". I'll merge this PR, and then I'll work on getting the underlying issue fixed so that we don't keep reverting to the wrong case. Thanks again for the PR, and for bringing this to our attention!

PR #894, opened 2020-12-28
Update aws-resource-elasticloadbalancingv2-listener.md

superdaigo:
The example code causes an error due to the value's type mismatch.
The type of the `Port` value is "Integer" instead of "String".

*Issue #, if available:*
I haven't checked the issue # since it's just a kind of typo.

*Description of changes:*
The type of the `Port` value is Integer instead of String. The integer value should not be surrounded by quotations in YAML.
I've removed the double quotations (") around the `443`.


rachfop:
Thank you for your request. A member of the team will be working on it and will provide updates here about progress.

ddarbari (approved):

rachfop:

PR #896, opened 2020-12-30
Update aws-properties-s3-websiteconfiguration.md

itshili:
*Issue #, if available:* 
If the resources AWS::S3::Bucket WebsiteConfiguration is defined without IndexDocument propert the stack will fail with the error below.

> S3Bucket CREATE_FAILED Property IndexDocument cannot be empty

*Description of changes:*
The IndexDocument should be set to required yes.

The following S3 documentation seem to suggest that as well:
https://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteHosting.html
https://docs.aws.amazon.com/AmazonS3/latest/dev/EnableWebsiteHosting.html


sasake615:
The "Type", "Required", "and "Update behavior" fields for CFN properties are auto-generated from the resource specification. I've submitted a ticket to correct this value in the resource spec. The source documentation will be updated after the specification is updated.

PR #898, opened 2021-01-06
Update the SourceAccessConfiguration property to include up-to-date properties

lewgordon:
*Description of changes:*

Adding additional valid properties for https://docs.aws.amazon.com/lambda/latest/dg/API_SourceAccessConfiguration.html.  During my testing today it seems like they should work now.


lewgordon:
I didn't update the description as I wasn't sure what would be required to change there.  It's probably worth noting for a few of the listed types the lambda function will need additional permissions.

lewgordon:
Just noticed this is gone now @rachfop and @devlinbd2 , although I do think my documentation was still valid.  Do you know if I should have edited some other file?

alivest:
@lewgordon Hello! Sorry about that, your PR is absolutely correct. I'll get it fixed and the weekly refresh cadence will pick up the changes.

lewgordon:
@alivest no problem.  Thank you for the quick response!

PR #904, opened 2021-01-13
correcting update behavior descriptions

PatMyron:
[`AWS::CloudTrail::Trail.TrailName`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudtrail-trail.html#cfn-cloudtrail-trail-trailname) `requires: Replacement`

inaccurate to translate `"UpdateType": "Conditional"` from the [Resource Specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) to say it `retains the physical ID` as [detailed here](https://github.com/aws-cloudformation/cloudformation-cli/pull/367#discussion_r395955803)

[`replacementStrategy`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/86) can differ from the default


rachfop (approved):
CloudTrail.Trail is interesting. 
1. Can we substitute an example in which the sentence "For example, if you update any property on the [Resource Name] resource, AWS CloudFormation updates the [resource] without disruption"? 
2. "a version property might be upgradable in-place but not downgradable in-place" - acknowledged
3. replacementStrategy - acknowledged

PatMyron:
> 1. Can we substitute an example in which the sentence "For example, if you update any property on the [Resource Name] resource, AWS CloudFormation updates the [resource] without disruption"?

Definitely other resource types where that statement would still be true, but any new property type in the future could invalidate that statement (imagine that's what happened with [`AWS::CloudTrail::Trail.TrailName`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudtrail-trail.html#cfn-cloudtrail-trail-trailname)), so I figured the wording I put would be more future-proof

PR #908, opened 2021-01-17
Update aws-resource-cloudtrail-trail.md

karimelmel:
*Description of changes:*
Fixed the example stack by adding IncludeGlobalServiceEvents to True, a Multi-region trail will fail to deploy if that is not enabled hence the stack does currently not work if deployed as in the documentation. 


swiercek:

PR #914, opened 2021-01-26
fixed incorrect FileSystemPolicy principal

jessedye:
The file system policy was incorrect and needed "AWS:" added to it. Otherwise this creates a 400 error when you try to execute the cloudformation.

*Issue #, if available:*

*Description of changes:* 

Fixes issue with incorrect Principal for FileSystemPolicy. The example policy needs to have "AWS:" to reference an account in AWS and will not work from a copy/paste perspective.

I agree


rachfop:
Thank you for your suggestion. A member of our team will use this thread to communicate any updates.

PR #917, opened 2021-01-29
Update aws-resource-lambda-layerversionpermission.md

cchostak:
This is not clear in the docs, but just received a case where this message came up -> [cfn-lint] W3011: Both UpdateReplacePolicy and DeletionPolicy are needed to protect Resources/LayerPermissions from deletion

Would be great to have it in the docs as well

*Issue #, if available:*

*Description of changes:*


alivest:
Looks good, thank you!

PR #918, opened 2021-01-29
AWS::Transfer::Server - Clarify that `Ref` does not return multiple values, just ARN.

sturdiva:
Previous docs seemed to indicate multiple values were returned via `Ref`, when in fact it is just the ARN. 


rachfop:
Hey, @sturdiva, just following up on this. I have submitted a ticket internally to our doc team to have this pull request looked at. Thank you for your feedback it is appreciated.

PR #920, opened 2021-02-03
Update using-cfn-get-template.md

cchostak:
Added a clarification in the docs as cli do not perform the same steps as the Management console does.

*Issue #, if available:*

*Description of changes:*


rachfop:
Hello @ccostak. Thank you for taking the time to submit feedback to our docs. Every piece of feedback is appreciated. 
The current doc information and your changes are both correct. The **Retrieving a template** page is nested under the **Using the AWS CLI**, so information on how the console behaves _isn't_ needed. However, does knowing the differences between the CLI and the console behavior when retrieving a template something you've encountered? Thanks!

PR #922, opened 2021-02-09
feat: Add example of how to set a custom SessionDuration to a SSO Per…

AaronKalair:
*Issue #, if available:* N/A

*Description of changes:* I was confused about the exact valid way to specify a `SessionDuration` so I added an example to clarify.


AaronKalair:
Hey,

Anything I can do to help move this along? :)

rachfop:
Thank you for the ping and for your submission. There is no further action required from your side.

PR #960, opened 2021-04-08
The title/desc for 2 return values were transposed

thegareth:
*Description of changes:*

In the return value section, the description of of the IsDefault/VpcId values were transposed. I swapped the descriptions so that they described the correct item


thegareth:
Looks like this was updated already

PR #966, opened 2021-04-22
Add note on how to use `ResourceTypeList`

xiaket:
*Issue #, if available:*

It is unclear as to how to use `ResourceTypeList` in the template.

*Description of changes:*

Add note on this otherwise vague way to apply a policy to multiple resource types.


judypear:
Thank you for adding this!

PR #967, opened 2021-04-22
Fix requirement for Vcpus and Memory

christiam:
This patch is to clarify that the `Vcpus` and `Memory` fields in the JobDefinition's ContainerProperties are required for the EC2 platform, but not for Fargate.

### DIFF
--- a/doc_source/aws-properties-batch-jobdefinition-containerproperties.md
+++ b/doc_source/aws-properties-batch-jobdefinition-containerproperties.md
@@ -142,7 +142,7 @@ This parameter indicates the memory hard limit \(in MiB\) for a container\. If y
 This parameter maps to `Memory` in the [Create a container](https://docs.docker.com/engine/api/v1.23/#create-a-container) section of the [Docker Remote API](https://docs.docker.com/engine/api/v1.23/) and the `--memory` option to [docker run](https://docs.docker.com/engine/reference/run/)\.  
 This parameter is supported on EC2 resources but isn't supported on Fargate resources\. For Fargate resources, you should specify the memory requirement using `resourceRequirement`\. You can also do this for EC2 resources\.  
 If you're trying to maximize your resource utilization by providing your jobs as much memory as possible for a particular instance type, see [Memory Management](https://docs.aws.amazon.com/batch/latest/userguide/memory-management.html) in the *AWS Batch User Guide*\.
-*Required*: No  
+*Required*: Yes (for EC2 Platform), No (for Fargate)
 *Type*: Integer  
 *Update requires*: [No interruption](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-no-interrupt)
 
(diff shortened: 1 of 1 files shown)


rachfop:
@christiam I'm letting Lee review this one, but the accepted values for Required are: `Yes`, `No`, and `Conditional`. 

We can add a note for the conditions. Thanks for your feedback!

HLeeHart:
This suggestion is not accurate. On EC2, it's not necessary to specify the Vcpu or Memory requirements using [ContainerProperties](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-containerproperties.html). Vcpu and Memory requirements can also be specified using [ResourceRequirement](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-resourcerequirement.html).

The Memory and Vcpu docs already say this.

christiam:
@HLeeHart , 
      Thanks for the feedback, but I have that not to be the case. Below is an error message I got when the Memory requirement was not provided:

`An error occurred (ClientException) when calling the RegisterJobDefinition operation: Error executing request, Exception : Memory is not provided., RequestId: 17bb1254-3039-4b29-91a6-56fe494ba74d`

FWIW, it seems that others have also encountered this issue: https://stackoverflow.com/questions/63817847/register-job-definition-in-aws-batch-giving-registerjobdefinition-operation-er

I hope this helps, regards,

Christiam

HLeeHart:
Can you show me what your ResourceRequirement looks like? It should look something like this:

```
 "ResourceRequirements" : [ 
  {
    "Type" : "MEMORY",
    "Value" : "4096"
  }, 
  {
    "Type" : "VCPU",
    "Value" : "1"
} ],
```

christiam:
Here's my job definition. If I omit `ContainerProperties.Vcpus` or `ContainerProperties.Memory`, I get the error I mentioned earlier. Why is that the case, if those fields are supposedly optional?

```yaml
  JobDefinition:
    Type: AWS::Batch::JobDefinition
    Properties:
      Type: container
      ContainerProperties:
        Image: !Ref Image
        Vcpus: 1
        Memory: 80000
        Command:
          - "run.sh"
          - "-i"
          - "Ref::input"
          - "-o"
          - "Ref::output"
          - "-b"
          - "Ref::batchlen"

christiam:
@HLeeHart , @rachfop , do you have any additional feedback? TIA!

HLeeHart:
Sorry, I was on vacation when you sent your response.
You must specify the Vcpu and Memory requirements. You can do this using either ContainerProperties.Memory and ContainerProperties.Vcpus or you can include ResourceRequirements. Can you show the job definition that has the ResourceRequirements specified?

christiam:
@HLeeHart , no problem, and thanks for getting back to me. My JobDefinition's ContainerProperties does not have a ResourceRequirements specified (as I understand it in the [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-containerproperties.html#cfn-batch-jobdefinition-containerproperties-resourcerequirements), it's not required).

PR #972, opened 2021-05-07
Update region in yaml, to match json example

prithvijj:
*Issue #, if available:*

*Description of changes:*
- Changing the region in the `yaml` example, to match the `json` example


rachfop:
Thank you for the feedback!

PR #977, opened 2021-05-13
Changed info for PrefferedBackupWindow on Replica Instances

mattbryce93:
*Description of changes:*

This Change adds a clarification for enabling backups on replica instances. Generally such a requirement would never be necessary, but it can sometimes be required. For example using the Replica instance as a DMS source endpoint, backups must be enabled to allow the binary log creation facilitating CDC migration modes. 


randyurbano:
Thanks for your feedback. The commit shows almost all of the content in the topic changing. Can you provide information on the specific parts that you want to update?

randyurbano:
We document the following for the SourceDBInstanceIdentifier property:

"Don't set the BackupRetentionPeriod, DBName, MasterUsername, MasterUserPassword, and PreferredBackupWindow properties. The database attributes are inherited from the source DB instance, and backups are disabled for Read Replicas."

Do you think this information is incorrect?

PR #984, opened 2021-05-19
Reformat the AppSync API cache example.

gredlerj:
*Issue #, if available:* n/a

*Description of changes:*
- Made the examples readable for both Yaml and JSON
- Corrected missing "Ref" for the reference to the example parameter "graphQlApiId"
- Corrected reference to T2_SMALL in the example.  Should not recommend against using the EC2-styled cache types in the property description then turn around and use them in an example.


KristinaPeterson:
Thank you for telling us about this issue. A member of the team will be working on it and will provide updates here about progress.

KristinaPeterson (approved):
Thank you again for your corrections to the formatting issues in this example.

PR #987, opened 2021-05-25
DLM Lifecycle example is missing the 'service-role' part of the ARN.

hal58th:
*Issue #, if available:*
None
*Description of changes:*
The example for lifecycle policy was missing the "service-role" part in the ARN. Without this, it does not work.


swiercek:

hal58th:
Already fixed

PR #989, opened 2021-05-26
chore: Fix inline example for VersionName

edogden-pbx:
*Issue #, if available:*
N/A

*Description of changes:*
Change the example for VersionName to match the regex pattern


JimbroAmazon:
Thanks Patrick!

PR #996, opened 2021-06-09
fix(blue-green.md): Scale value should be 100%, not 1%.

dacgray:
*Issue #, if available:*

Taskset should have scale percent of 100%, not 1% - since there's only one service in this example.

*Description of changes:*

- Update fargate version to latest
- change taskset scale value to be 100%


rachfop:
Thank you for your submission.

PR #1000, opened 2021-06-13
Fix invalid backup vault ARN

BenChaimberg:
Use "backup-vault" as ARN resource name instead of incorrect "vault".

---


rachfop:
Thank you for your feedback.

PR #1003, opened 2021-06-14
fix: MaxEntries maximum 100 to 1000 of prefix list

miyataka:
*Issue #, if available:*

*Description of changes:*
- because,
image
- And I can create prefix list which has 101 entry by cloudformation.

### DIFF
--- a/doc_source/aws-resource-ec2-prefixlist.md
+++ b/doc_source/aws-resource-ec2-prefixlist.md
@@ -50,7 +50,7 @@ Valid Values: `IPv4` \| `IPv6`
 One or more entries for the prefix list\.  
 *Required*: No  
 *Type*: List of [Entry](aws-properties-ec2-prefixlist-entry.md)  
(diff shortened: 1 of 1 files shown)


miyataka:
ping
Why this pull request is stalen?
Is there any missing information?

julieso:
In the console, "Max entries" is equivalent to the MaxEntries property, which is the total number of entries that you can add to a prefix list. It is this property that has a limit of 1,000; see https://docs.aws.amazon.com/vpc/latest/userguide/amazon-vpc-limits.html#vpc-quotas-managed-prefix-lists. The Entry property is where you add entries to the prefix list. The API limits you to adding 100 entries at a time: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateManagedPrefixList.html. So you can add entries in multiple batches, up to the point that you reach the maximum size of the prefix list that you configured.

PR #1013, opened 2021-06-26
Change principal to avoid errors if example is modified by user

shadycuz:
*Issue #, if available:*

*Description of changes:*
Added AWS key to principal policy. If a user copy+paste this policy and then change the principal to a specific role the template will show up as valid but will fail to deploy because of a "Invalid policy syntax" error that is very hard to track down. Thats because when using the "Principal" key directly only a wildcard is valid. 

See `Anonymous user principals (public access)` section in [AWS JSON policy elements: Principal](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html) documentation.

### DIFF
--- a/doc_source/aws-properties-s3-policy.md
+++ b/doc_source/aws-properties-s3-policy.md
@@ -89,7 +89,9 @@ The following sample is a bucket policy that is attached to the DOC\-EXAMPLE\-BU
                                 ]
                             ]
                         },
-                        "Principal": "*",
+                        "Principal": {
+                            "AWS": "*"
+                        },
                         "Condition": {
                             "StringLike": {
                                 "aws:Referer": [
@@ -123,7 +125,8 @@ SampleBucketPolicy:
             - - 'arn:aws:s3:::'
               - !Ref DOC-EXAMPLE-BUCKET
               - /*
-          Principal: '*'
+          Principal: 
+            AWS: '*'
           Condition:
             StringLike:
               'aws:Referer':


sasake615:
Thank you for your patience and for telling us about this issue. A member of the team will be working on it and will provide updates here about progress.

swiercek:

lincolahanbeck:
The [IAM documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html#principal-anonymous) states, 

> For anonymous users, the following elements are equivalent: 
`"Principal": "*"`
`"Principal": { "AWS" : "*" }`

In the [S3 docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteAccessPermissionsReqd.html), similar policy examples use `"Principal": "*"`. So we prefer to stick with the original option for consistency.

shadycuz:
@lincolahanbeck 

Maybe something changed, I haven't retested it but this was a huge issue when I reported it. 

If you take the current example:
```yml
SampleBucketPolicy:
  Type: 'AWS::S3::BucketPolicy'
  Properties:
    Bucket: !Ref DOC-EXAMPLE-BUCKET
    PolicyDocument:
      Statement:
        - Action:
            - 's3:GetObject'
          Effect: Allow
          Resource: !Join
            - ''
            - - 'arn:aws:s3:::'
              - !Ref DOC-EXAMPLE-BUCKET
              - /*
          Principal: '*'
          Condition:
            StringLike:
              'aws:Referer':
                - 'http://www.example.com/*'
                - 'http://example.net/*'
```

and you change the principal to an ARN...

```yml
SampleBucketPolicy:
  Type: 'AWS::S3::BucketPolicy'
  Properties:
    Bucket: !Ref DOC-EXAMPLE-BUCKET
    PolicyDocument:
      Statement:
        - Action:
            - 's3:GetObject'
          Effect: Allow
          Resource: !Join
            - ''
            - - 'arn:aws:s3:::'
              - !Ref DOC-EXAMPLE-BUCKET
              - /*
          Principal: 'arn:aws:iam::account:role/my-ec2-instance-profile'
          Condition:
            StringLike:
              'aws:Referer':
                - 'http://www.example.com/*'
                - 'http://example.net/*'
```

This code would pass Cloudformation "verify" but would fail to actually deploy a stack. This is an extremely hard to debug error.

I understand that other service documentation uses the "shorter" form and says they are equivalent, but when I opened this issue they were not equivalent with Cloudformation.

PR #1018, opened 2021-07-08
Change code snippet default value for InstanceTenancy

woodharter:
The dedicated value in InstanceTenancy caused me a few hours of delay after causing a very vague error message in a cloudformation system I was starting to build. I finally found this solution https://stackoverflow.com/questions/45691826/aws-launch-configuration-error-the-requested-configuration-is-currently-not-sup and back traced the problem to this piece of documentation.  I think the most common reason someone would copy this code is a new build and unlikely that a newb would want dedicated tenancy. I think it would be best to set this flag to default so others don't get sent down this same rabbit hole.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/doc_source/aws-resource-ec2-vpc.md
+++ b/doc_source/aws-resource-ec2-vpc.md
@@ -117,10 +117,11 @@ The following example specifies a VPC\.
    "Type" : "AWS::EC2::VPC",
    "Properties" : {
       "CidrBlock" : "10.0.0.0/16",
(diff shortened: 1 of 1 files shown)


sasake615:
Thank you for your patience and for telling us about this issue. A member of the team will be working on it and will provide updates here about progress.

julieso:
Thanks for the suggestion. I agree that it is better to show a more typical configuration.

PR #1021, opened 2021-07-10
Update Welcome.md

blange:
On line #7 "normally" doesn't exactly match up with the word "might".  "Normally" describes the most common scenario while "might" describes a possible scenario. "Would" is a better choice here. I removed "would" from the subsequent sentence to avoid repetition. 

Thank you for taking the time to review my proposed change. 

-Brian

*Issue #, if available:*

*Description of changes:*


rachfop:
Thank you for your feedback!

PR #1051, opened 2021-08-30
Update the pattern of Security Group Arns

paulkilla:
The pattern is completely wrong, and references a File System rather than the EC2 Security Group.


draliatp-amz:
Hi! Thanks for pointing this out. We found one other place that this pattern incorrect so I'm going to go ahead and make the change in both places.

PR #1056, opened 2021-09-07
Remove stray space and superfluous newlines

siebrand:


judypear:
Hi Siebrand! 
Thanks so much for that stray space correction! 
As for the superfluous newlines, those are autogenerated by the build process that creates these .md files. I'll accept your changes to those, but they'll be overwritten by the next build that comes through. 
I've incorporated your bugfix for the stray space - I really appreciate your help keeping these docs in good shape.
-Judy

PR #1059, opened 2021-09-16
Fix the 172.16 CIDR to be correct

timothybonci:
*Issue #, if available:*

*Description of changes:*
The current CIDR in the docs is invalid, and should be updated to remove the typo.


jimdial-aws:
Thank you for your submission and for using Amazon EKS.

PR #1065, opened 2021-09-22
remove incorrect optional in AWS::SecretsManager::Secret

jogold:
*Description of changes:*
It is required so it cannot be optional...


lizcraw:
Thanks for catching this!

PR #1068, opened 2021-09-24
Fix missing "Ref" in JSON for cost anomaly detection example.

sturdiva:
YAML was correct, JSON was not. AWS::CE::AnomalySubscription already
had the Ref's as well.

*Issue #, if available:*

*Description of changes:*


amiito:
Thank you for your pull request. Your change has been merged in the Git Hub doc, and will be reflected in the CFN user guide (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ce-anomalymonitor.html) shortly to match.

PR #1093, opened 2021-11-16
Fixed Type field in YAML example for AWS::CertificateManager::Certificate

t04glovern:
YAML example for AWS::CertificateManager::Certificate has the Type field in the incorrect location.


rachfop:
Thank you for the fix!

PR #1107, opened 2021-12-16
Removed duplicate line in TaskType definition

carstenosu:
There were two 'Allowed Values' lines in the TaskType documentation. I removed the second one which was also out of date. The first one contains the correct list of allowed values.

*Issue #, if available:*

*Description of changes:*
Removed duplicate AllowedValues line


rachfop:
Thank you for the fix!

PR #1112, opened 2021-12-21
Fix RDS Database Instance LicenseModel valid values

matt-schwartz:
*Issue #, if available:* N/A

*Description of changes:*

Value values listed for RDS instance LicenseModel was inaccurate.  It's dependent on the selected engine.  It's also not easy to find the available options for each engine, so I've expanded on it in this CF documentation.


randyurbano:
Thanks for your feedback. The commit seems to update a large amount of unrelated content, so I'll update the documentation without merging this commit.

matt-schwartz:
Sounds good.  The rest are newline changes which I couldn't seem to avoid.  It's got mixed CR and CRLF newlines I believe.  Feel free to update separately and close this PR.  Thanks!

PR #1173, opened 2022-03-01
fix: appsync caching example syntax

vicary:


KristinaPeterson:
Thank you. I'm investigating.

PR #1180, opened 2022-03-10
fix: remove repeated properties

GlennChia:
*Issue #, if available:*

*Description of changes:*

Properties are repeated in the example


mark-amzn:
Also fixed in source files; same fix applied for JSON version of the example.

PR #1213, opened 2022-04-29
fix secret description to match service

cat5inthecradle:
*Description of changes:*

Description mentions RDS, but actually is being used for Redshift in this instance.


cat5inthecradle:
upstream change removed the RDS reference.

PR #1223, opened 2022-05-25
Improved description for Engagement

dosatos:
*Issue #, if available:*
The documentation is not accurate, saying it should be a string.
I tried name and alias and it did not work. I spent a couple of hours debugging until found out that had to be an "arn".

*Description of changes:*
Improved the description to specify it should be an "arn".


JimbroAmazon:
Published updated description for this field, which states:

The Amazon Resource Name (ARN) for the contacts and escalation plans that the response plan engages during an incident.

This is the appropriate fix for this issue.

